PR #1123, opened 2014-08-06
Make migration process a bit more solid

vuntz:
We validate proposals on migration, and also give more information about the script where a crash might occur.


toabctl:
Handled by https://github.com/crowbar/barclamp-crowbar/pull/1271 now.

PR #1131, opened 2014-09-05
Test scaffolding

mapleoin:
I'm trying to setup some test scaffolding in order to encourage more tests to be written both to the `crowbar_framework/` and `chef/*` code. When this goes through, I'd like to add similar code to all of our barclamps.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,17 @@
 language: ruby
-rvm:
-  - "1.8.7"
-branches:
-  only:
-    - master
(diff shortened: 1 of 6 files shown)


tboerger:
So this won't be used that way anymore.

vuntz:
@tboerger what will be the new way? :-)

tboerger:
@vuntz As everything is running on the same ruby version now we don't need most of that stuff. We just need to add chefspec without the bash scripts and test matrix and so on.

PR #1146, opened 2014-10-15
Node is not getting roles during discovery

ovisan:
fixing https://bugzilla.novell.com/show_bug.cgi?id=900216
FATAL -- :Node exists without role!!

### DIFF
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -192,7 +192,8 @@ class NodeObject < ChefObject
       end
     end
     # deep clone of @role.default_attributes, used when saving node
-    @attrs_last_saved = deep_clone(@role.default_attributes)
+    # fixing https://bugzilla.novell.com/show_bug.cgi?id=900216 with unless @role.nil?
+    @attrs_last_saved = deep_clone(@role.default_attributes) unless @role.nil?
     @node = node
   end
 


tboerger:
Can we simply merge this pull request and create another bsc entry to fix that properly in the next release? With this pull request the UI stays intact even after forgetting a node.

aspiers:
@tboerger I'm OK with that but only if the bsc entry gets created **before** this gets merged, so that it can't get forgotten ;-)

ovisan:
@aspiers https://bugzilla.novell.com/show_bug.cgi?id=915154 please see if this fits the need (else i can change it accordingly)

aspiers:
@ovisan Thanks that will do for now - we can change the title and add other details later :)

ovisan:
@aspiers great! :-)

vuntz:
Did anyone see that bug recently? Because I'm really unsure that the patch won't break stuff later on (ie, push back the bug to a later stage).

tboerger:
We have seen that a short while ago at my installation and at ovisans installation.

ovisan:
i saw it again yesterday...

aspiers:
@tboerger @ovisan supportconfigs or it didn't happen ;-)

dirkmueller:
I think this might be very risky.. and hiding the real bug elsewhere. I don't see an explanation on why this is the right fix. -1 until we have better understood it

ovisan:
OK, for me is fine, i can modify this file anytime i get this error. Not sure is so easy for the others who are using it.

tboerger:
So just try to forget a node and see it on your own. I'm tired of such a discussion. somebody serves a fix, only complains about a dirty fix, nobody brings in a better fix. So just keep this damn thing like it is and live with bug reports from customers the crowbar does not work anymore.

aspiers:
> Because I'm really unsure that the patch won't break stuff later on (ie, push back the bug to a later stage).

@vuntz That's true, but even then it's still not a good reason to avoid merging this.  The UI should be fixed so that it's robust in the face of backend issues.  If we merge this and it uncovers other UI weaknesses, then we are exchanging one crash for others, but that's *good* because then we can fix them too.  I don't think anyone is proposing to sweep the underlying issue under the carpet, and if we reopen [bsc#915154](https://bugzilla.suse.com/show_bug.cgi?id=915154) then we can track that separately.

vuntz:
@aspiers it's in the model, so it's not just about UI. This could create severe issues in the backend.

aspiers:
True, so what I originally proposed in https://github.com/crowbar/barclamp-crowbar/pull/1146#issuecomment-60370768 is probably the best way to go.

vuntz:
I'm nearly sure that the right fix is something like https://github.com/crowbar/barclamp-crowbar/pull/1233

bkutil:
Agreed with @vuntz - I think we should close this PR and focus the effort on #1233 , which makes more sense (is stricter and very vocal about the invalid input, so the issues above won't occur)

PR #1149, opened 2014-10-26
Cleanup ProposalObject#status

aspiers:
I'm really tired so maybe my brain is missing something, but these look like obvious refactorings to some dumb code.


vuntz:
Still needs to be rebased.

tboerger:
@aspiers please rebase, after that we can merge that i think :)

aspiers:
Rebased.

tboerger:
:+1:

PR #1152, opened 2014-11-10
new crowbar_batch script

aspiers:
***This needs the `easy_diff` gem!  If there are issues with introducing that to Cloud 4, maybe this should not be merged.***

(backport of #1147 to `release/stoney/master`)

    crowbar batch build foo.yaml

will automatically create (if necessary), edit, and apply all proposals described within `foo.yaml`.  This enables building arbitrary cloud configurations in an entirely automated fashion, which can be very useful for testing, building reproducible demos, and eventually maybe even partially mimicking customer environments.


aspiers:
@vuntz What was the conclusion of this?

aspiers:
@vuntz ping again!

vuntz:
@aspiers can you sync the code with what we have in master? There were some fixes there, if I'm not mistaken. I'll try to get the missing gem available for Cloud 4.

aspiers:
@vuntz Done.  I included some changes to `barclamp_lib.rb` which were probably not strictly necessary, but thought it would be cleaner this way if we need to do any future syncing.

aspiers:
Oh yeah, I didn't test yet :-)

vuntz:
So I realized recently that one issue we might have (packaging-wise) is that Cloud 4 is ruby 1.8 and Cloud 5 is ruby 2.1, with different package naming guidelines :/ Still, should be possible.

tboerger:
LGTM codewise, have not tested it.

aspiers:
@vuntz Why would that be an issue?  Are you talking about the packaging of `easy_diff` or something else?

vuntz:
@aspiers yes, talking about the `easy_diff` packaging. It would have been easier (in terms of process) if we could just have taken the Cloud 5 package on Cloud 4.

aspiers:
Wait - isn't this already done?

https://build.suse.de/package/show/Devel:Cloud:Shared:11-SP3/rubygem-easy_diff

vs.

https://build.suse.de/package/show/Devel:Cloud:Shared:Rubygem/rubygem-easy_diff

vuntz:
@aspiers it's not about creating the package, it's about getting it in the update channel :-) It's easier if it's already in another product.

aspiers:
OK, I'll leave you IBS / product wizards to worry about that then ;-)

PR #1161, opened 2014-12-05
Make replaceSpace replace all whitespaces, not just space

mjura:


tboerger:
@mjura Explain it better in commit message please :)

rsalevsky:
:+1: works great, from my side the commit message is clear enough

vuntz:
@rsalevsky maybe you can clarify why it's needed, then? :-) I'm still unsure why this is the case.

rsalevsky:
`\s` matches every whitespace including tabs and linebreaks. Another improvement here is we are independent from the encoding.

vuntz:
I understand the `\s` bit. I'm just not sure that we should also replace `\s*,\s*` as there's no explanation of why it's needed and it changes the function to do more than what its name tells us.

mjura:
@vuntz  : The problem is that when somebody will provide data in following format `string1, string2, string3`, then this will be transformed into `string1,,string2,,string3` which makes validation error for attributes.

vuntz:
@mjura: surely it's a bug in splitString, then, not in this function.

mjura:
@vuntz : In my opinion this regexp should be fixed here, because it parses fields and prepares it for splitString function where `,` are only used as delimiters. But I don't insist, it is only proposal of problem which I have found.

vuntz:
It really depends if "foo,,,,bar" should be split to ["foo", "", "", "", "bar"] or ["foo", "bar"]. I think it should be the latter, hence me thinking we should fix splitString. But if the former is what we expect, then your fix is right.

aspiers:
> It really depends if "foo,,,,bar" should be split to ["foo", "", "", "", "bar"] or ["foo", "bar"].

FWIW, Ruby, Python, and Perl all return the former, and that makes more sense to me, e.g. think of parsing CSV records with empty fields.

vuntz:
>> It really depends if "foo,,,,bar" should be split to ["foo", "", "", "", "bar"] or ["foo", "bar"].
>
>FWIW, Ruby, Python, and Perl all return the former, and that makes more sense to me, e.g. think of parsing CSV records with empty fields.

In a general context, yes. But I'm not sure we care about empty elements in the context of this function here.

Anyway, I guess I'll leave it up to others to share their feelings on this. I'm happy being overruled here.

(I still believe the commit message can be clearer, though ;-))

vuntz:
> (I still believe the commit message can be clearer, though ;-))

Just to be a bit more constructive, I'd split the commit in two commits, and use these messages:

```
Make replaceSpace replace all whitespaces, not just space
```

and

```
Change replaceSpace to also take into accounts commas

We do not want "foo, bar" to be split in ["foo", "", "bar"] but ["foo", "bar"].
```

mjura:
Added new commit and comments changed

bkutil:
+1

rsalevsky:
still :+1:

tboerger:
Got no strong opinion on that anymore. +1

mjura:
Thank you guys :)

PR #1162, opened 2014-12-05
delete _remove roles from the run_list

MaximilianMeister:
When a `$barclamp-$role_remove` role (and its removal recipe) has successfully finished, delete it from the node's run_list.


vuntz:
In general, this looks good. But one concern I had a looong time ago and that I just remembered: if the chef-client fails and apply_role ends before we reach this code, I'm afraid that we will have nodes that are not known to the proposal anymore but that still have the _remove role. How will they lost this role now?

MaximilianMeister:
@vunz maybe that should be handled with a proper exception [here](https://github.com/MaximilianMeister/barclamp-crowbar/blob/delete_remove_role/crowbar_framework/app/models/service_object.rb#L1330) as a separate issue?

tboerger:
What's the status of this? @vuntz, @bkutil: Anything left open?

vuntz:
> @vuntz maybe that should be handled with a proper exception [here](https://github.com/MaximilianMeister/barclamp-crowbar/blob/delete_remove_role/crowbar_framework/app/models/service_object.rb#L1330) as a separate issue?

Hrm, not sure what you mean?

The way I would possibly do it is save somewhere in the databag the info that we're removing the roles from some nodes at the beginning, and once we reach the end of the method, I would remove that info from the databag.

> What's the status of this? @vuntz, @bkutil: Anything left open?

Yes, see my last comment ;-)

MaximilianMeister:
i accidentally removed my branch, moved to https://github.com/crowbar/barclamp-crowbar/pull/1224

PR #1166, opened 2014-12-17
Fix changing the platform in bulk view

rhafer:
Selecting a non-default target platform for nodes in the bulk view didn't have
any effect if "allocate" was marked at the same time. Handling the target
platform attribute before the allocate attribute should fix the problem.


vuntz:
+1

PR #1167, opened 2015-01-05
Bnc863101

bkutil:
A fix for  https://bugzilla.suse.com/show_bug.cgi?id=863101

Applying a barclamp allocates (i.e. re-installs) a node.  This could be
problematic e.g. if the node contains a previously installed OS, or user
data.

This patch changes the policy to not allocate the nodes automatically on
barclamp apply.

This is preferrable to not offering the non-allocated nodes in the UI as
it works also for the API. User will be informed about the need to
allocate the nodes manually.

### DIFF
--- a/crowbar_framework/app/controllers/barclamp_controller.rb
+++ b/crowbar_framework/app/controllers/barclamp_controller.rb
@@ -391,10 +391,16 @@ class BarclampController < ApplicationController
           flash[:notice] = answer[1] if answer[0] >= 300 and answer[0] < 400
           flash[:notice] = t('barclamp.proposal_show.commit_proposal_success') if answer[0] == 200
           if answer[0] == 202
-            flash_msg = answer[1].map { |node_dns|
-                 NodeObject.find_node_by_name(node_dns).alias
-            }.join ", "
-            flash[:notice] = "#{t('barclamp.proposal_show.commit_proposal_queued')}: #{flash_msg}"
+            missing_nodes = answer[1].map { |node_dns| NodeObject.find_node_by_name(node_dns) }
+
+            unready_nodes = missing_nodes.map(&:alias)
+            unallocated_nodes = missing_nodes.reject(&:allocated).map(&:alias)
+
+            flash[:notice] = "#{t('barclamp.proposal_show.commit_proposal_queued')}: #{(unready_nodes - unallocated_nodes).join(", ")}"
+
+            unless unallocated_nodes.empty?
+              flash[:alert]  = "#{t('barclamp.proposal_show.commit_proposal_queued_unallocated')}: #{unallocated_nodes.join(", ")}"
+            end
           end
         rescue StandardError => e
           flash_and_log_exception(e)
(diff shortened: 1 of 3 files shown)


crowbar_framework/app/models/service_object.rb
```diff
@@ -219,8 +219,7 @@ def add_pending_elements(bc, inst, elements, queue_me, pre_cached_nodes = {})
           next if node.nil?
           pre_cached_nodes[node_name] = node
 
-          # Make sure the node is allocated
-          node.allocate!
+          # Mark node as pending. User will be informed about manual allocation
```
vuntz: Can you add a "if node is not allocated" at the end of this comment? Otherwise, it's a bit confusing.

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -391,10 +391,16 @@ def proposal_update
...
           flash[:notice] = t('barclamp.proposal_show.commit_proposal_success') if answer[0] == 200
           if answer[0] == 202
-            flash_msg = answer[1].map { |node_dns|
-                 NodeObject.find_node_by_name(node_dns).alias
-            }.join ", "
-            flash[:notice] = "#{t('barclamp.proposal_show.commit_proposal_queued')}: #{flash_msg}"
+            missing_nodes = answer[1].map { |node_dns| NodeObject.find_node_by_name(node_dns) }
+
+            unready_nodes = missing_nodes.map(&:alias).join(", ")
+            unallocated_nodes = missing_nodes.reject(&:allocated).map(&:alias).join(", ")
```
vuntz: Do I understand right that an unallocated node will appear in the two messages? Shouldn't we have them in the unallocated message only?

bkutil: I'm not sure. Won't it be confusing to say 'hey, it is unallocated, but I'm not waiting for it to be ready' ?

crowbar_framework/app/models/service_object.rb
bkutil: Sure, excuse my bat Engrish.

crowbar_framework/app/controllers/barclamp_controller.rb
bkutil: Although I see your point in duplicating the information being bad.

vuntz: Yeah, I'm mostly thinking that the duplication is what could (possibly) be confusing (because of too much information). I don't have a very strong feeling about this, though. So if anyone else feels the current way is fine, we can just keep it.

bkutil: I thought about it some more and I think you're right. There is no point in duplicating the info.

vuntz:
+1

dirkmueller:
can you rewrite the commit subjects to "xxx (bnc#12345)" instead ? this is the more standardized format (suffix instead of prefix and bnc#)

bkutil:
@dirkmueller done

dirkmueller:
@bkutil thanks!

dirkmueller:
+1

PR #1169, opened 2015-01-08
Revert "Added redis and sidekiq gems"

dirkmueller:
Seems unused.

This reverts commit 5adbca8fe0a293cf252862e84d725c6577c74a84.


tboerger:
I have added them to add serversent events and background processing. So it is already planned to use them. So :-1: from me!

tboerger:
Note: Anyway you have removed it only from a single place instead of 2 places.

tboerger:
I want to keep that to have already all dependencies so that nobody can complain about dependencies that have been added too late in release cycle. I hope i can start to implement the functionality for that soonish

tboerger:
Ok just to avoid maintainance for tex.

PR #1170, opened 2015-01-08
Do not use deep munge

vuntz:
This is breaking the JSON parsing for empty arrays, which we use with
the magic crowbar-deep-merge-template feature when creating the initial
crowbar proposal.

With deep munge enabled, we cannot remove the nagios proposal, which
results in failures when nagios barclamp is not available.

See:
https://github.com/rails/rails/issues/8832
https://github.com/rails/rails/issues/13420

We can revert this once switching to rails 5 (see comments in second
issue).

### DIFF
--- a/crowbar_framework/config/application.rb
+++ b/crowbar_framework/config/application.rb
@@ -27,6 +27,8 @@ module Crowbar
 
     config.time_zone = "UTC"
 
+    config.action_dispatch.perform_deep_munge = false
+
     config.i18n.enforce_available_locales = true
     config.i18n.default_locale = :en
 


tboerger:
I would name it simply ```dispatch.rb```and you are missing the copyright header

vuntz:
I added the copyright header. However, I think the current name is appropriate since it's for `ActionDispatch`.

tboerger:
You are simply doing it on the complete wrong place :)

Please add ```config.action_dispatch.perform_deep_munge = false``` to https://github.com/crowbar/barclamp-crowbar/blob/master/crowbar_framework/config/application.rb instead of adding a initializer file.

tboerger:
+2

dirkmueller:
seems to fix the issue, +1

tboerger:
> We can revert this once switching to rails 5 (see comments in second
> issue).

Rails 5 will require one more ruby update. Rails 5 will have a hard dependency on Ruby 2.2 :D

PR #1171, opened 2015-01-08
Unbreak crowbar_batch with rails4 port (bnc#912309)

vuntz:
Looks like the replies are different now...

http://bugzilla.suse.com/show_bug.cgi?id=912309

### DIFF
--- a/bin/crowbar_batch
+++ b/bin/crowbar_batch
@@ -512,14 +512,14 @@ def export_proposal(barclamp, template, proposal, alias_by_host)
   to_wipe = squash(removed_attrs) - squash(added_attrs)
   unless to_wipe.empty?
     puts '  wipe_attributes:'
-    puts to_yaml(to_wipe).gsub(/^/, '    ')
+    puts batch_to_yaml(to_wipe).gsub(/^/, '    ')
   end
 
   if added_attrs.empty?
     puts '  attributes:'
   else
     attrs = { 'attributes' => added_attrs }
-    puts to_yaml(attrs).gsub(/^/, '  ')
+    puts batch_to_yaml(attrs).gsub(/^/, '  ')
   end
 
   elements = json['deployment'][barclamp]['elements']
@@ -536,16 +536,16 @@ def export_proposal(barclamp, template, proposal, alias_by_host)
       'elements' => elements
     }
   }
-  puts to_yaml(deployment).gsub(/^/, '  ')
+  puts batch_to_yaml(deployment).gsub(/^/, '  ')
 end
 
-def to_yaml(data)
+def batch_to_yaml(data)
   data.to_yaml.sub('---', '').strip.
     gsub(ALIAS_REGEXP, '\1')
 end
 
 def print_yaml(data)
-  puts(to_yaml(data))
+  puts(batch_to_yaml(data))
 end
 
 def main
--- a/bin/crowbar_batch
+++ b/bin/crowbar_batch
@@ -186,7 +186,7 @@ end
 def ensure_proposal_exists(barclamp, name)
   out, err = with_barclamp(barclamp) { proposal_list }
   abort out if err > 0
-  if ! out.find { |line| line.start_with? name }
+  if ! out.split("\n").find { |line| line.start_with? name }
     create_proposal(barclamp, name)
   else
     proposal_puts "Already exists"
--- a/bin/crowbar_batch
+++ b/bin/crowbar_batch
@@ -425,7 +425,7 @@ def export(barclamps)
     if code != 200
       abort "Failed to receive list of barclamps: #{code}: #{out}"
     end
-    barclamps = out.sort_by { |x| x[1]['order'] }.map { |x| x[0] }
+    barclamps = out['barclamp'].sort_by { |x| x[1]['order'] }.map { |x| x[0] }
   end
 
   puts '---'
@@ -458,7 +458,7 @@ def export_barclamp(barclamp, alias_by_host)
     end
   end
 
-  proposals = get_proposal_list(barclamp)
+  proposals = get_proposal_list(barclamp)[barclamp]
   if proposals.empty?
     puts "# WARNING: no proposals exist for #{barclamp} barclamp"
     return
--- a/bin/barclamp_lib.rb
+++ b/bin/barclamp_lib.rb
@@ -208,11 +208,11 @@ def list
 
   if struct[1] != 200
     [ "Failed to talk to service list: #{struct[1]}: #{struct[0]}", 1 ]
-  elsif struct[0].nil? or struct[0].empty?
+  elsif struct[0][@barclamp].nil? or struct[0][@barclamp].empty?
     [ "No current configurations", 0 ]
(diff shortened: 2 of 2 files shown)


bin/crowbar_batch
```diff
@@ -458,7 +458,7 @@ def export_barclamp(barclamp, alias_by_host)
     end
   end
 
-  proposals = get_proposal_list(barclamp)
+  proposals = get_proposal_list(barclamp)[barclamp]
```
tboerger: That feels wrong for me

vuntz: Why? Try it :-)

tboerger: Yeah, but calling it with barclamp and getting a barclamp entry afterwards?

vuntz:
FWIW, I now hit this:
```
crowbar batch build foo
[19:54:44] crowbar barclamp, 'default' proposal:
   Already exists
/usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/mixin/deep_merge.rb:207:in `deep_merge!': Use RbConfig instead of obsolete and deprecated Config.
/usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/mixin/deep_merge.rb:207: warning: toplevel constant Config referenced by Chef::Config
/usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/mixin/deep_merge.rb:207:in `deep_merge!': Use RbConfig instead of obsolete and deprecated Config.
/opt/dell/bin/barclamp_lib.rb:536:in `eval': undefined method `[]' for RbConfig:Module (NoMethodError)
        from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/mixin/deep_merge.rb:207:in `deep_merge!'
        from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/mixin/deep_merge.rb:138:in `block in deep_merge!'
        from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/mixin/deep_merge.rb:133:in `each'
        from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/mixin/deep_merge.rb:133:in `deep_merge!'
        from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/mixin/deep_merge.rb:138:in `block in deep_merge!'
        from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/mixin/deep_merge.rb:133:in `each'
        from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/mixin/deep_merge.rb:133:in `deep_merge!'
        from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/mixin/deep_merge.rb:138:in `block in deep_merge!'
        from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/mixin/deep_merge.rb:133:in `each'
        from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/mixin/deep_merge.rb:133:in `deep_merge!'
        from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/mixin/deep_merge.rb:138:in `block in deep_merge!'
        from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/mixin/deep_merge.rb:133:in `each'
        from /usr/lib64/ruby/gems/2.1.0/gems/chef-10.32.2/lib/chef/mixin/deep_merge.rb:133:in `deep_merge!'
        from ./crowbar_batch:330:in `merge_attributes'
        from ./crowbar_batch:249:in `modify_proposal'
        from ./crowbar_batch:182:in `build_proposal'
        from ./crowbar_batch:112:in `block in build'
        from ./crowbar_batch:111:in `each'
        from ./crowbar_batch:111:in `build'
        from (eval):1:in `run_sub_command'
        from /opt/dell/bin/barclamp_lib.rb:536:in `eval'
        from /opt/dell/bin/barclamp_lib.rb:536:in `run_sub_command'
        from /opt/dell/bin/barclamp_lib.rb:540:in `run_command'
        from ./crowbar_batch:553:in `main'
        from ./crowbar_batch:556:in `<main>'
```

bin/crowbar_batch
vuntz: I'm not the one who changed the API ;-) What is returned now is something like `["crowbar", ["default"]]`

vuntz:
Ok, seems to work now.

bin/crowbar_batch
tboerger: Yeah but this looks like an unwanted change of the API. Currently without accessing my codebase i have no idea where this have been introduced.

tboerger:
IMHO this is the wrong workaround. It looks like an unwanted API change.

bkutil:
So what seems to have happened here was, that the routes got changed (mea maxima culpa as I was responsible for fixing them after the transition), and then we observed the issue with our tools failing and fixed them.

So now we have two possible approaches:

1) taken by this patch is to fix the tools and keep the API as it is. As long as we're sure all our tools now work and that customers don't need cloud4 vs cloud5 compatibility between tools and the API, we're OK.
2) I go and fix the API, i.e. write a test that does the same request as the crowbar batch here and make sure that the code behaves the same under rails 2 and rails 4. Then this PR would be not needed and we're done (unless we'll have to revert some other PRs that were made to fix this in the meantime)

@vuntz @tboerger what do you think about this?

vuntz:
@bkutil I don't have any opinion on what's best, tbh.

FWIW, you can also see the API change by doing `crowbar $barclamp list` and other similar commands. So it's not just `crowbar batch`.

tboerger:
I prefer to fix the API

aspiers:
Right - `crowbar batch` doesn't contain any API code itself, it just reuses the same code that all the other CLI commands use.  So whatever approach you decide, please check they all still work ;-)

bkutil:
@tboerger the API / routes are unchanged, however, rails renders JSON with a root. A simple fix might be just to pas `:root => false` to the `render` calls, but there might be a global option (I was unable to find it after googling for some time, and ActiveRecord::Base.include_root_in_json does not seem to have an effect here).

tboerger:
It must be cause by something different. https://github.com/crowbar/barclamp-crowbar/blob/master/crowbar_framework/config/initializers/wrap.rb

bkutil:
@tboerger you can check for yourself, if you're willing to spend 10 minutes on it - copy the line from inside the block setting the activerecord::base option before the render in e.g. index action in barclamps controller, restart, then curl that endpoint. The setting will have no effect (at least that is what I observed). Maybe because what we're trying to render is not an activerecord base object?

vuntz:
Moved to https://github.com/crowbar/barclamp-crowbar/issues/1181 and https://github.com/crowbar/barclamp-crowbar/issues/1185

PR #1172, opened 2015-01-09
Fix performance issue w/ slow UI responses

bkutil:
This fixes the performance issues caused by missing HTTP response headers.

Also included is a patch that reduces the app server workers startup time, effectively reducing the time-to-first-response.

### DIFF
--- a/chef/cookbooks/crowbar/files/default/config.ru
+++ b/chef/cookbooks/crowbar/files/default/config.ru
@@ -16,4 +16,6 @@
 #
 
 require ::File.expand_path("../config/environment",  __FILE__)
+use Rack::Chunked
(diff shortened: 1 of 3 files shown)


vuntz:
I'll trust you on this, +1 :-)

bkutil:
@vuntz you don't have to - you can test for yourself https://github.com/bkutil/slow-rainbows

vuntz:
Note that you need to change https://github.com/crowbar/barclamp-crowbar/blob/master/chef/cookbooks/crowbar/files/default/config.ru too

vuntz:
+1

dirkmueller:
+1

PR #1173, opened 2015-01-12
Revert "Preload app code on rainbows startup"

bkutil:
This reverts commit 93a7e299ba41e3c59d444eafe8371ac7fc89bac3.

Our deployment scripts assume that force-reload is in fact equivalent to
restart (i.e. reloads the code), which is not the case if we preload the
app.


dirkmueller:
Quickly tested it, fixes the regression, +1

PR #1174, opened 2015-01-15
Automatically redirect to dashboard start page

dirkmueller:
Instead of showing a stupid error page, try to be helpful
and immediately go to the right page.

### DIFF
--- a/crowbar_framework/index/index.html
+++ b/crowbar_framework/index/index.html
@@ -110,6 +110,7 @@
       for (i = 0; i < links_to_dashboard.length; i++) {
         links_to_dashboard[i].href="http://" + window.location.host + ":3000";
       }
+      window.location.replace("http://" + window.location.host + ":3000");
     </script>
   </body>
 </html>


tboerger:
:-1: This have been handled all the time that way. Maybe we have to think about proxying the apache to the rails app?

tboerger:
Any other opinion on that? Maybe a good alternative should be a redirect in apache?

vuntz:
+1

crowbar_framework/index/index.html
```diff
@@ -115,6 +115,7 @@
 
     <script>
       document.getElementById("dashboard").href="http://" + window.location.host + ":3000";
+      window.location.replace(document.getElementById("dashboard"));
```
vuntz: (not that it matters much, but I'd use a temporary variable to store the URL and use it on the two lines)

bkutil:
@dirkmueller You could also drop most of the page and use https://en.wikipedia.org/wiki/Meta_refresh#Examples - would that make sense?

cc @vuntz

tboerger:
As @bkutil said, just use ```<meta http-equiv="refresh" content="0; url=http://example.com/">```, but keep the page as it is for people who have deactivated javascript and stuff like that. It would be better to always present a real page :)

bkutil:
@tboerger I realized that the problem with the html-based redirects is with the hard-coded URL. I think we'd have to make sure a correct IP/domain is in the HTML, while javascript based redirect will take the URL from the browser, and just change that.

bkutil:
+1

tboerger:
I would suggest another approach

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <!-- snip -->
  </head>
  <body>
    <div class="navbar navbar-inverse navbar-fixed-top" role="navigation">
      <div class="container">
        <div class="navbar-header">
          <button class="navbar-toggle" data-target=".navbar-collapse" data-toggle="collapse">
            <span class="sr-only">
              Navigation
            </span>
            <span class="icon-bar"></span>
            <span class="icon-bar"></span>
            <span class="icon-bar"></span>
          </button>
          <a href="/" class="navbar-brand">
            <span class="title">Crowbar</span>
          </a>
        </div>
        <div class="collapse navbar-collapse navbar-right">
          <ul class="nav navbar-nav">
            <li><a id="dashboard" href="http://127.0.0.1:3000/">Go to dashboard</a></li>
          </ul>
        </div>
      </div>
    </div>

    <div class="container" id="errorpage">
      <div class="panel panel-default panel-warning">
        <div class="panel-heading">
          <h2>
            Crowbar Admin Server on port 3000
          </h2>
        </div>
        <div class="panel-body">
          <p>
            The Crowbar Admin Server is only running on port 3000, so you will be redirected in a few seconds. If
            that doesn't happen please follow the link at the top menu to the Crowbar dashboard. If you still have
            problems feel free to visit the <a href="http://github.com/crowbar/crowbar/wiki">Crowbar site</a> for
            more information.
          </p>
        </div>
      </div>

      <footer>
        <span>
          Provided by <a href="http://dell.com/cloud" target="_blank">Dell CloudEdge Solutions team</a>
        </span>
      </footer>
    </div>

    <script>
      document.getElementById("dashboard").href = [
        "http://",
        window.location.host,
        ":3000"
      ].join("")

      setTimeout(
        function() {
          window.location.replace(document.getElementById("dashboard"));
        },
        5000
      );
    </script>
  </body>
</html>

rsalevsky:
@tboerger Please rebase

tboerger:
What about this pull request? More opinions?

rsalevsky:
We should replace the current setup with a Puma, Passenger or any other correct ruby server setup instead of doing a workaround.

tboerger:
> We should replace the current setup with a Puma, Passenger or any other correct ruby server setup instead of doing a workaround.

This is unrelated. This is the page that gets displayed on port 80 where we have an apache running. This page should be just anotice that the user wants to switch to port 3000.

rsalevsky:
> This is unrelated. This is the page that gets displayed on port 80 where we have an apache running. This page should be just anotice that the user wants to switch to port 3000.

Right, I mean we should fix this with using Puma + Apache Proxy. So also -1 from my side.

tboerger:
Obsoletes with #1325

dirkmueller:
closed in favor of https://github.com/crowbar/barclamp-crowbar/pull/1325

PR #1175, opened 2015-01-15
Add two new validation constraints.

aplanas:
* Add "platform" constraint:

  "platform" => ['suse', /12.(0|1)/]

  This will expect that the service have a platform equal to 'suse'
  and a platform_version of 12.0 or 12.1

* Add "exclude_platform" constraint:

  "exclude_platform" => [/suse/, '11.3']

  If the platform is a SLES11 SP3, the service will not be installed.

### DIFF
--- a/crowbar_framework/app/assets/javascripts/jquery/nodeList.js
+++ b/crowbar_framework/app/assets/javascripts/jquery/nodeList.js
@@ -70,6 +70,8 @@
             node,
             source.data('alias'),
             source.data('admin'),
+            source.data('platform'),
+            source.data('platform-version'),
             source.data('cluster'),
             true
           );
@@ -121,6 +123,8 @@
           $node.data('id'),
           $node.data('alias'),
           $node.data('admin'),
+          $node.data('platform'),
+          $node.data('platform-version'),
           $node.data('cluster'),
           false
         );
@@ -203,7 +207,7 @@
     );
   };
 
-  NodeList.prototype.insertNode = function(role, id, alias, admin, cluster, initial) {
+  NodeList.prototype.insertNode = function(role, id, alias, admin, platform, platform_version, cluster, initial) {
     var self = this;
     var $role = $(self.dataBag.roleTarget.format(role));
 
@@ -317,6 +321,40 @@
             break;
         }
       }
+
+      if (constraints.platform !== undefined && constraints.platform) {
+        var platforms = $.map(constraints.platform, function(c_version, c_platform) {
+          return (platform === c_platform) && self.equalOrMatchVersion(platform_version, c_version);
+        });
+        var is_any_true = platforms.some(function (element, index, array) {
+          return element;
+        });
+        if (!is_any_true) {
+          return self.errorMessage(
+            'barclamp.node_selector.platform'.localize().format(
+              alias,
+              role
+            )
+          );
+        }
+      }
+
+      if (constraints.exclude_platform !== undefined && constraints.exclude_platform) {
+        var platforms = $.map(constraints.exclude_platform, function(c_version, c_platform) {
+          return (platform === c_platform) && self.equalOrMatchVersion(platform_version, c_version);
+        });
+        var is_any_true = platforms.some(function (element, index, array) {
+          return element;
+        });
+        if (is_any_true) {
+          return self.errorMessage(
+            'barclamp.node_selector.exclude_platform'.localize().format(
+              alias,
+              role
+            )
+          );
+        }
+      }
     }
 
     $role.append(
@@ -357,6 +395,66 @@
     );
 
     this.dataBag.removedOld = true;
+  };
+
+  NodeList.prototype.equalOrMatchVersion = function(value, maybe_regexp) {
+    var to_version = function(version) {
+      var match = /^(\d+)\.(\d+)\.(\d+)$/.exec(version);
+      if (match) {
+        return [parseInt(match[1]), parseInt(match[2]), parseInt(match[3])];
+      }
+
+      match = /^(\d+)\.(\d+)$/.exec(version);
+      if (match) {
+        return [parseInt(match[1]), parseInt(match[2]), 0];
+      }
+
+      match = /^(\d+)$/.exec(version);
+      if (match) {
+        return [parseInt(match[1]), 0, 0];
+      }
+
+      return [0, 0, 0];
+    };
+
+    var cmp = function(op, version1, version2) {
+      var as_version1 = to_version(version1);
+      var as_version2 = to_version(version2);
+      var result = as_version1.map(function(v1, i) {
+        switch (op) {
+          case ">=":
+            return v1 >= as_version2[i];
+          case ">":
+            return v1 > as_version2[i];
+          case "<=":
+            return v1 <= as_version2[i];
+          case "<":
+            return v1 < as_version2[i];
+          case "==":
+            return v1 == as_version2[i];
+          default:
+            return false;
+        }
+      });
+
+      return result.every(function(element, index, array) {
+        return element;
+      });
+    };
+
+    // First group contains the RegExp without pre and post '/'
+    var is_re = /^\/(.*)\/$/.exec(maybe_regexp);
+    if (is_re) {
+      var re = new RegExp(is_re[1]);
+      return re.test(value)
+    }
+
+    var op_value = /^(>=?|<=?)\s*([.\d]+)$/.exec(maybe_regexp);
+    if (op_value) {
+      return cmp(op_value[1], value, op_value[2]);
+    } else {
+      return cmp('==', value, maybe_regexp);
+    }
   };
 
   $.fn.nodeList = function() {
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -855,6 +855,36 @@ class ServiceObject
     false
   end
 
+  def violates_platform_constraint?(elements, role)
+    if role_constraints[role] && role_constraints[role].has_key?("platform")
+      constraints = role_constraints[role]["platform"]
+      elements[role].each do |element|
+        next if is_cluster? element
+        node = NodeObject.find_node_by_name(element)
+
+        return true if !constraints.any? do |platform, version|
+          node[:platform] == platform && equal_or_match_version(node[:platform_version], version)
+        end
+      end
+    end
+    false
+  end
+
+  def violates_exclude_platform_constraint?(elements, role)
+    if role_constraints[role] && role_constraints[role].has_key?("exclude_platform")
+      constraints = role_constraints[role]["exclude_platform"]
+      elements[role].each do |element|
+        next if is_cluster? element
+        node = NodeObject.find_node_by_name(element)
+
+        return true if constraints.any? do |platform, version|
+          node[:platform] == platform && equal_or_match_version(node[:platform_version], version)
+        end
+      end
+    end
+    false
+  end
+
   def violates_cluster_constraint?(elements, role)
     if role_constraints[role] && !role_constraints[role]["cluster"]
       clusters = elements[role].select {|e| is_cluster? e}
@@ -892,6 +922,16 @@ class ServiceObject
       if violates_admin_constraint?(elements, role, nodes_is_admin)
         validation_error("Role #{role} does not accept admin nodes.")
         break
+      end
+
+      if violates_platform_constraint?(elements, role)
+        platforms = role_constraints[role]["platform"].map {|k, v| [k, v].join(' ')}.join(', ')
+        validation_error("Role #{role} can be used only for #{platforms} platform(s).")
+      end
+
+      if violates_exclude_platform_constraint?(elements, role)
+        platforms = role_constraints[role]["exclude_platform"].map {|k, v| [k, v].join(' ')}.join(', ')
+        validation_error("Role #{role} can't be used for #{platforms} platform(s).")
       end
 
       if violates_cluster_constraint?(elements, role)
@@ -1574,6 +1614,37 @@ class ServiceObject
       Rack::MiniProfiler.step(name, &block)
     else
       block.call
+    end
+  end
+
+  def equal_or_match_version(value, maybe_regexp)
+    to_version = lambda do |version|
+      case version.to_s
+      when /^(\d+)\.(\d+)\.(\d+)$/
+        [ $1.to_i, $2.to_i, $3.to_i ]
+      when /^(\d+)\.(\d+)$/
+        [ $1.to_i, $2.to_i, 0 ]
+      when /^(\d+)$/
+        [ $1.to_i, 0, 0 ]
+      else
+        [0, 0, 0]
+      end
+    end
+
+    cmp = lambda do |op, version|
+      (to_version.call(value).zip to_version.call(version)).all? do |v1, v2|
+        v1.send(op, v2)
+      end
+    end
+
+    op_value = maybe_regexp.scan(/^(>=?|<=?)\s*([.\d]+)$/)
+
+    if maybe_regexp.start_with?('/') && maybe_regexp.end_with?('/')
+      Regexp.new(maybe_regexp[1..-2]).match(value)
+    elsif op_value.length > 0
+      cmp.call(op_value.first.first, op_value.first.last)
+    else
+      cmp.call('==', maybe_regexp)
     end
   end
 end
--- a/crowbar_framework/app/views/barclamp/_node_selector.html.haml
+++ b/crowbar_framework/app/views/barclamp/_node_selector.html.haml
@@ -49,7 +49,7 @@
         .form-group
           %ul.nodes
             - nodes_hash.each do |n, detail|
-              %li.form-control.input-sm.available{ "data-draggable" => "true", "data-id" => n, "data-alias" => detail[:alias], "data-cluster" => "false", "data-admin" => detail[:admin].to_s, :title => (detail[:title] || t("not_set")) }
+              %li.form-control.input-sm.available{ "data-draggable" => "true", "data-id" => n, "data-alias" => detail[:alias], "data-cluster" => "false", "data-admin" => detail[:admin].to_s, "data-platform" => detail[:platform], "data-platform-version" => detail[:platform_version], :title => (detail[:title] || t("not_set")) }
                 = icon_tag :hdd
                 = detail[:alias]
                 = link_to icon_tag("link"), node_path(detail[:handle]), :title => t(".node_link", :alias => detail[:alias]), :class => "pull-right"
--- a/crowbar_framework/spec/models/service_object_spec.rb
+++ b/crowbar_framework/spec/models/service_object_spec.rb
@@ -131,5 +131,91 @@ describe ServiceObject do
         dns_service.validation_errors.first.should match(/cannot be assigned to both role/)
       end
     end
+
+    describe "platform" do
+      it "allows nodes of matched platform using operators" do
+        dns_service.stubs(:role_constraints).returns(
+          {
+            "dns-client" => {
+              "admin" => true ,
+              "platform" => { "ubuntu" => ">= 10" }
+            }
+          })
+        dns_service.validate_proposal_constraints(dns_proposal)
+        dns_service.validation_errors.length.should be == 0
+      end
+
+      it "allows nodes of matched platform with fancy versioning" do
+        dns_service.stubs(:role_constraints).returns(
+          {
+            "dns-client" => {
+              "admin" => true ,
+              "platform" => { "ubuntu" => "10.10.0" }
+            }
+          })
+        dns_service.validate_proposal_constraints(dns_proposal)
+        dns_service.validation_errors.length.should be == 0
+      end
+
+      it "allows nodes of matched platform using regular expressions" do
+        dns_service.stubs(:role_constraints).returns(
+          {
+            "dns-client" => {
+              "admin" => true ,
+              "platform" => { "ubuntu" => "/10.*/" }
+            }
+          })
+        dns_service.validate_proposal_constraints(dns_proposal)
+        dns_service.validation_errors.length.should be == 0
+      end
+
+      it "allows nodes of matched platform using regular expressions (multiple platforms)" do
+        dns_service.stubs(:role_constraints).returns(
+          {
+            "dns-client" => {
+              "admin" => true ,
+              "platform" => { "suse" => "12.0", "ubuntu" => "/10.*/" }
+            }
+          })
+        dns_service.validate_proposal_constraints(dns_proposal)
+        dns_service.validation_errors.length.should be == 0
+      end
+
+      it "does not allow nodes of a different platform" do
+        dns_service.stubs(:role_constraints).returns(
+          {
+            "dns-client" => {
+              "admin" => true ,
+              "platform" => { "suse" => "12.0" }
+            }
+          })
+        dns_service.validate_proposal_constraints(dns_proposal)
+        dns_service.validation_errors.first.should match(/can be used only for suse 12.0/)
+      end
+
+      it "does not allow nodes of a different platform (multiple parforms)" do
+        dns_service.stubs(:role_constraints).returns(
+          {
+            "dns-client" => {
+              "admin" => true ,
+              "platform" => { "suse" => "12.0", "redhat" => "/.*/" }
+            }
+          })
+        dns_service.validate_proposal_constraints(dns_proposal)
+        dns_service.validation_errors.first.should match(/can be used only for suse 12.0/)
+      end
+
+      it "does not allow nodes of a Ubuntu" do
+        dns_service.stubs(:role_constraints).returns(
+          {
+            "dns-client" => {
+              "admin" => true ,
+              "exclude_platform" => { "ubuntu" => "/.*/" }
+            }
+          })
+        dns_service.validate_proposal_constraints(dns_proposal)
+        dns_service.validation_errors.first.should match(/can't be used for ubuntu/)
+      end
+    end
   end
 end
--- a/crowbar_framework/app/assets/javascripts/misc/stringLocalize.js
+++ b/crowbar_framework/app/assets/javascripts/misc/stringLocalize.js
@@ -29,6 +29,8 @@ if (!String.prototype.localize) {
       'barclamp.node_selector.unique': 'Failed to assign {0} to {1}, it\'s already assigned to another role',
       'barclamp.node_selector.zero': 'Failed to assign {0} to {1}, no assignment allowed',
       'barclamp.node_selector.max_count': 'Failed to assign {0} to {1}, maximum of allowed nodes/clusters reached',
+      'barclamp.node_selector.platform': 'Failed to assign {0} to {1}, this platform is not allowed',
+      'barclamp.node_selector.exclude_platform': 'Failed to assign {0} to {1}, this platform is excluded',
       'barclamp.node_selector.conflicting_roles': 'Node {0} cannot be assigned to both {1} and any of these roles: {2}'
     };
 
--- a/crowbar_framework/app/helpers/nodes_helper.rb
+++ b/crowbar_framework/app/helpers/nodes_helper.rb
@@ -94,6 +94,8 @@ module NodesHelper
             :title => node.description(false, true),
             :admin => node.admin?,
             :group => node.group,
+            :platform => node[:platform],
+            :platform_version => node[:platform_version],
             :cluster => false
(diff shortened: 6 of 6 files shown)


crowbar_framework/app/models/service_object.rb
```diff
@@ -854,6 +854,56 @@ def violates_admin_constraint?(elements, role, nodes_is_admin = {})
...
+      elements[role].each do |element|
+        next if is_cluster? element
+        node = NodeObject.find_node_by_name(element)
+
+        (platform, version) = role_constraints[role]["platform"]
+        platform = {
+          :platform => platform,
+          :platform_version => version
+        }
+        platform.any? do |key, value|
```
jsuchome: I think you need to explictly return the result of this (also platform.all? in the second function)

crowbar_framework/app/models/service_object.rb
```diff
@@ -854,6 +854,56 @@ def violates_admin_constraint?(elements, role, nodes_is_admin = {})
     false
   end
 
+  def violates_platform_constraint?(elements, role)
+    if role_constraints[role]
```
jsuchome: Please add a check that the constraint is present: ``&& role_constraints[role].has_key?("platform")``

aplanas:
* Return the result of all? and any?
* Skip the evaluation if the role_constraint do not have the "platform" key

tboerger:
:-1: The client side validation is missing

aplanas:
@tboerger Where is the client side validation code?

crowbar_framework/app/models/service_object.rb
```diff
@@ -893,6 +943,28 @@ def validate_proposal_constraints(proposal)
         break
       end
 
+      if violates_platform_constraint?(elements, role)
```
tboerger: Why is this part of the cluster constraint check?

tboerger:
I would prefer a hash for the constraint definition. The key is the platform and the value can be the specific version constrai t in form of a string, regexp or integer/float.

tboerger:
You can find the javascript for clientside galidation at https://github.com/crowbar/barclamp-crowbar/blob/master/crowbar_framework/app/assets/javascripts/jquery/nodeList.js#L237

vuntz:
@aplanas I'm not sure there's a way to express something like the following with your patch: "if platform is SLE, then SLE12 only; accept all non-SLE platforms"

jsuchome:
@vuntz I think for that, we have "exclude_platform" => [/suse/, '11.3']

vuntz:
@jsuchome well, it's "if platform is SLE, then not SLE11-SP3", which is slightly different than "if platfirm is SLE, then SLE12 only". But it might be good enough for now, indeed.

tboerger:
But the pattern is still messy

aplanas:
Yep, I think that "Accept SLES12 or anything that is not SLE" is approx. to "Exclude SLES11*" for this use case.

As @jsuchome this is something like { "exclude_platform" => ['suse', /9|(1[01].*)/] }

crowbar_framework/app/models/service_object.rb
aplanas: It is not. I guess that you are tricked by the github collapse algorithm.

tboerger: You are right

aplanas:
I guess that to alleviate the messy regexp we can use @tboerger approach of using hashes instead of list.

```ruby
  {
      "exclude_platform" => {
        "suse": "9",
        "suse": "10.*",
        "suse": "11.*"
      }
  }
```
But the semantic is the same : m

tboerger:
That won't work. But regular expressions are fine. You can even negate the value of the regex. But hash keys have to be unique.

aplanas:
Uau. Sometimes I amuse myself : (

Of course the hash is wrong. But we can use some lists here:

```ruby
  {
      "exclude_platform" => [
        ["suse", "9"],
        ["suse", "10.*"],
        ["suse", "11.*"]
      ]
  }
```

tboerger:
And why do you strictly rely on arrays? Something like this is IMHO absolutly fine. What do you think @bkutil?

```ruby
{
  "exclude_platform" => {
    "suse" => /[^12]/
  }
}
```

jsuchome:
I agree with use of hashes as @tboerger proposes. At least we do not have to convert arrays to hashes in validate functions

aplanas:
I am not sure that I like the hash solution, are we going to allow this?

```ruby
{
  "exclude_platform" => {
    "suse" => /[^12]/,
    "debian" => /.*/
  }
}
```

tboerger:
Of course is that allowed.

aplanas:
* Change list with hashes

Add two new validation constraints.
    
* Add "platform" constraint:

```ruby
"platform" => {
  {
    'suse' => /12.(0|1)/,
    /debian/ => /.*/
  }
}
```

This will expect that the service have a platform equal to 'suse' and a platform_version of 12.0 or 12.1, or a debian.

* Add "exclude_platform" constraint:

```ruby
"exclude_platform" => {
  {
    'suse' => '11.3',
     /redhat/ => /.*/
  }
}
```

If the platform is a SLES11 SP3 or a redhat, the service will not be installed.

aplanas:
Uhm, if I add regular expression, the javascript serializer do not work properly

tboerger:
Please dont use regex for the hash keys.

aplanas:
* Add JS validators
* Use strings to encapsulate regular expressions in JSON
* Disable both constraints when cluster

```
    Add two new validation constraints.
    
    * Add "platform" constraint:
    
      "platform" => {
         {
           'suse' => '/12.(0|1)/',
           'debian' => '/.*/'
         }
      }
    
      This will expect that the service have a platform equal to 'suse'
      and a platform_version of 12.0 or 12.1, or a debian.
    
    * Add "exclude_platform" constraint:
    
      "exclude_platform" => {
        {
           'suse' => '11.3',
           'redhat' => '/.*/'
        }
      }
    
      If the platform is a SLES11 SP3 or a redhat, the service will not
      be installed.
    
    * Add constrains to JavaScript validator
    
      Because JSON can't contains regular expressions, the constraints
      needs to be expressed as strings.  If the string starts and ends
      with '/', the JavaScript and the Ruby part takes care of convert
      it into a Regexp.
```

crowbar_framework/app/models/service_object.rb
```diff
@@ -893,6 +923,16 @@ def validate_proposal_constraints(proposal)
...
+        platforms = role_constraints[role]["platform"].map {|k, v| [k, v].join(' ')}.join(', ')
+        validation_error("Role #{role} can be used only for #{platforms} platform(s).")
+      end
+
+      if violates_exclude_platform_constraint?(elements, role)
+        platforms = role_constraints[role]["platform"].map {|k, v| [k, v].join(' ')}.join(', ')
```
jsuchome: "exclude_platform"

aplanas:
* Fix error message for exclude_platform constraint

crowbar_framework/app/models/service_object.rb
```diff
@@ -893,6 +923,16 @@ def validate_proposal_constraints(proposal)
...
       end
 
+      if violates_platform_constraint?(elements, role)
+        platforms = role_constraints[role]["platform"].map {|k, v| [k, v].join(' ')}.join(', ')
+        validation_error("Role #{role} can be used only for #{platforms} platform(s).")
+      end
+
+      if violates_exclude_platform_constraint?(elements, role)
+        platforms = role_constraints[role]["exclude_platform"].map {|k, v| [k, v].join(' ')}.join(', ')
+        validation_error("Role #{role} can't be used in #{platforms} platform(s).")
```
jsuchome: Shouldn't there be 'for xy platforms' in this message as well?

aplanas:
* Fix another typo in the error message
* Rebase

crowbar_framework/app/assets/javascripts/jquery/nodeList.js
```diff
@@ -359,6 +397,16 @@
     this.dataBag.removedOld = true;
   };
 
+  var equal_or_match = function(value, maybe_regexp) {
```
tboerger: Please dont pullute the namespace. Make it as ```NodeList.prototype.equalOrMatch``` and call it with ```self.equalOrMatch()```

crowbar_framework/app/models/service_object.rb
```diff
@@ -1575,5 +1615,12 @@ def profile(name, &block)
     else
       block.call
     end
+
```
tboerger: ```def profile``` is missing the closing ```end```

crowbar_framework/app/views/barclamp/_node_selector.html.haml
```diff
@@ -35,7 +35,7 @@
             %ul.clusters
               - @service_object.available_clusters.sort.each do |cluster_name, role|
                 - cluster_barclamp, cluster_proposal = @service_object.cluster_get_barclamp_and_proposal(cluster_name)
-                %li.form-control.input-sm.available{ "data-draggable" => "true", "data-id" => cluster_name, "data-alias" => cluster_proposal, "data-cluster" => "true", "data-admin" => "false", :title => cluster_proposal }
+                %li.form-control.input-sm.available{ "data-draggable" => "true", "data-id" => cluster_name, "data-alias" => cluster_proposal, "data-cluster" => "true", "data-admin" => "false", "data-platform" => "cluster", "data-platform-version" => "cluster", :title => cluster_proposal }
```
tboerger: I think this is wrong. That way we can not do real checks on that. Maybe display a list of platforms?

aplanas:
* Fix wrong rebase

aplanas:
* Do not polute the JS namespace: convert equal_or_match into a prototype method of NodeList class

crowbar_framework/app/views/barclamp/_node_selector.html.haml
aplanas: IDK what is correct here. I excluded both constraints for clusters, is this OK?

tboerger: I think we have to check the platform for clusters as well.

tboerger:
Except the cluster constraint check it looks fine so far.

crowbar_framework/app/views/barclamp/_node_selector.html.haml
jsuchome: In general, I agree, but currently we skip all validation checks for cluster, so it is not specific to this new constraint.

vuntz: The difference here is that we won't skip the check, but do it with wrong values.

aplanas: @vuntz Most possibly I am missing something, but the check is skip in Ruby code (https://github.com/crowbar/barclamp-crowbar/pull/1175/files#diff-0be8ea1f05c9441a59bb2b7ac65b1c62R861) and in JavaScript code (https://github.com/crowbar/barclamp-crowbar/pull/1175/files#diff-0be8ea1f05c9441a59bb2b7ac65b1c62R861)

vuntz: Oh, sorry, you're right. Do we really need to set these (wrong) attributes, then?

tboerger: In that case i would skip that attribute. But for the future it should be interesting to check that as well.

aplanas: IDK, is added into the NodeList insertNode method, and I provide and isomorphic data structure for both clustered and single node.  Maybe an empty value, instead of a random one like 'cluster' is more correct here?

tboerger: Just skip the attribute? It should only match if the attribute is defined.

aplanas:
* Remove data-platform and data-platform-version from the cluster (now the data-* is different from single nodes configuration)

dirkmueller:
I wonder if this should be really an "exclude" based constraint and not an "only_if" (e.g. include only) constraint. I'd say the opt-in approach is more reusable (if we really want to go in and support install of non-sles platforms which we just removed during the rails4 port)

aplanas:
@dirkmueller Maybe I didn't understand your comment, but the constraint is both: exclude and "only_if". 

If you want to exclude SLE12 you can use the exclude constraint: ```exclude_platform => {'suse' => '12.0'}```, but if you only can install the barclamp in SLE12, you need to use the ```platform => {'suse' => '12.0'}``` one.

dirkmueller:
@aplanas Thanks for the explanation. If I read the patch correctly, you can set "exclude_platform" and "platform" at the same time, and if any of those match then it will reject it. I was wondering if they should be somehow combineable, so that you can do things like: 

exclude_platform {'rhel': any'},
platform{'rhel', 'rhel7'}

which refuses all rhel version other than rhel7. in the current implementation it will reject basicall all possible platforms I think.

crowbar_framework/app/assets/javascripts/jquery/nodeList.js
```diff
@@ -317,6 +321,40 @@
             break;
         }
       }
+
+      if (constraints.platform !== undefined && constraints.platform && !cluster) {
```
tboerger: No need for the cluster check as it doesn't have a platform constraint attribute.

crowbar_framework/app/assets/javascripts/jquery/nodeList.js
```diff
@@ -317,6 +321,40 @@
...
+          return self.errorMessage(
+            'barclamp.node_selector.platform'.localize().format(
+              alias,
+              role
+            )
+          );
+        }
+      }
+
+      if (constraints.exclude_platform !== undefined && constraints.exclude_platform && !cluster) {
```
tboerger: No need for the cluster check as it doesn't have a platform exclude constraint attribute.

tboerger:
@dirkmueller Combining the constraints in that way won't work with the current implementation.

aplanas:
@dirkmueller Sure. This was discussed in this same thread. As I understand your use case is "If RHEL7 or other thing different from RHEL install, else do not install".  This similar to this: ```'exclude_platform' => {'rhel' => '/rhel[56]/'}```

The solutions that I think to combine both are not very intuitive. Your example needs to discover that one match is more specific that another, and this is not easy when are regular expressions involved.

jsuchome:
I think we do not need to overengineer this, I think current proposal is good enough

tboerger:
> I think we do not need to overengineer this, I think current proposal is good enough

:+1:

vuntz:
Just checking something: do we have a way to say "the only platform that works is suse-11.3" (so non-suse doesn't work, and suse != 11.3 doesn't work either). We basically need this for the pacemaker barclamp right now...

tboerger:
Sure

"platform" => { "suse" => "11.3" }

aplanas:
@vuntz: Is the same case that Ceph: https://github.com/crowbar/barclamp-ceph/pull/110

vuntz:
@tboerger @aplanas no, that's not the same. I understand `"platform" => { "suse" => "11.3" }` as "only 11.3 for suse, but anything for non-suse". And that's fine for ceph, but that's different from "only 11.3 and only for suse".

aplanas:
@vuntz: I tryed to express in the code that ```"platform" => { "suse" => "11.3" }``` means "install only in SLE11.3" and ```"exclude_platform" => { "suse" => "11.3" }``` means "install in every platform except SLE11.3"

If you want to express "install in every platform, but if is SUSE, only install if 11.3" was more or less equivalent to "exclude SLE12"

tboerger:
@vuntz i think your opinion on that feels wired... If you take the unused crowbar.yml platform constraint or the chef metadata.yml platform constraints they are working exactly like that. if platform is defined run it only on these platforms.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1576,4 +1616,12 @@ def profile(name, &block)
       block.call
     end
   end
+
+  def equal_or_match(value, maybe_regexp)
```
tboerger: Make the constraints more flexible without regular expressions.

* /regexp/
* >=
* >
* <=
* <
* ==

```ruby
  def equal_or_match(value, maybe_regexp)
    case
    when maybe_regexp.start_with?('/') && maybe_regexp.end_with?('/')
      Regexp.new(maybe_regexp[1..-2]).match(value)
    when maybe_regexp.start_with?('>')
      op, check = maybe_regexp.split(" ", 2)
      value.to_f.send(op.to_sym, check.to_f)
    when maybe_regexp.start_with?('>=')
      op, check = maybe_regexp.split(" ", 2)
      value.to_f.send(op.to_sym, check.to_f)
    when maybe_regexp.start_with?('<')
      op, check = maybe_regexp.split(" ", 2)
      value.to_f.send(op.to_sym, check.to_f)
    when maybe_regexp.start_with?('<=')
      op, check = maybe_regexp.split(" ", 2)
      value.to_f.send(op.to_sym, check.to_f)
    else
      value == maybe_regexp
    end
  end
```

tboerger: Note: Needs to be more DRY

aplanas:
* Integrate operators (==, >=, >, <, <=) in ruby and javascript validators

* Add RSpec tests

crowbar_framework/app/models/service_object.rb
```diff
@@ -1576,4 +1616,28 @@ def profile(name, &block)
...
+    operator = lambda do
+      op, check = maybe_regexp.split(" ", 2)
+      value.to_f.send(op.to_sym, check.to_f)
+    end
+    
+    case
+    when maybe_regexp.start_with?('/') && maybe_regexp.end_with?('/')
+      Regexp.new(maybe_regexp[1..-2]).match(value)
+    when maybe_regexp.start_with?('==')
```
tboerger: I would skip that. If there is no operator defined at the beginning i would always take the equal check.

crowbar_framework/app/assets/javascripts/jquery/nodeList.js
```diff
@@ -359,6 +397,35 @@
...
+    if (is_re) {
+      var re = new RegExp(is_re[1]);
+      return re.test(value)
+    }
+
+    // Check if is an operator
+    var op_value = maybe_regexp.split(" ", 2);
+    if (op_value.length == 2) {
+      switch (op_value[0]) {
+        case "==":
```
tboerger: I would skip that. If there is no operator defined at the beginning i would always take the equal check. Just add a ```default``` section.

crowbar_framework/app/models/service_object.rb
aplanas: Is not exactly the same: "12.0" can match "12.0" and "== 12", but not "12"

crowbar_framework/app/assets/javascripts/jquery/nodeList.js
aplanas: Same rationale that in the previous comment: "== 12" to match "12.0"

crowbar_framework/app/models/service_object.rb
tboerger: Why? Both values are getting converted to a float. So it should match fine.

aplanas: If there is not operator, there is not conversion AFAICT. Is using a simple string comparison. If is prefixed with one operator (like '==') there is conversion to_f.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1576,4 +1616,28 @@ def profile(name, &block)
...
+    when maybe_regexp.start_with?('>=')
+      operator.call
+    when maybe_regexp.start_with?('>')
+      operator.call
+    when maybe_regexp.start_with?('<=')
+      operator.call
+    when maybe_regexp.start_with?('<')
+      operator.call
+    else
+      value == maybe_regexp
```
tboerger:
```ruby
if value =~ /\A[-+]?[0-9]+\z/
  value.to_f == maybe_regexp.to_f
else
  value == maybe_regexp
end
```

As addition to get rid of the ```==``` operator above.

tboerger: Even better one:

```ruby
if value.to_f.to_s == value
  value.to_f == maybe_regexp.to_f
else
  value == maybe_regexp
end
```

aplanas: I do not see the problem with '=='. Also many version like '10.10' do not behave correctly for the to_f.to_s implicit conversion.

What is the main point of removing the explicit '==' for numeric versions?  I would prefer not to remove this and avoid implicit conversions hidden from the developer.

aplanas:
* Remove '==' operator to follow chef trends.

jsuchome:
+1

crowbar_framework/app/models/service_object.rb
```diff
@@ -855,6 +855,36 @@ def violates_admin_constraint?(elements, role, nodes_is_admin = {})
...
+    false
+  end
+
+  def violates_exclude_platform_constraint?(elements, role)
+    if role_constraints[role] && role_constraints[role].has_key?("exclude_platform")
+      elements[role].each do |element|
+        next if is_cluster? element
+        node = NodeObject.find_node_by_name(element)
+
+        constraints = role_constraints[role]["exclude_platform"]
```
vuntz: This should be out of the loop.

aplanas:
* Remove superfluous check in JS code

crowbar_framework/app/models/service_object.rb
```diff
@@ -855,6 +855,36 @@ def violates_admin_constraint?(elements, role, nodes_is_admin = {})
...
+        return constraints.any? do |platform, version|
+          equal_or_match(node[:platform], platform) && equal_or_match(node[:platform_version], version)
+        end
```
vuntz: There's the question of sharing the core of these two methods in a small helper. For instance:

```ruby
def matches_platform_constraint?(elements, constraints)
  elements.each do |element|
    next if is_cluster? element
    node = NodeObject.find_node_by_name(element)

    return !constraints.any? do |platform, version|
      equal_or_match(node[:platform], platform) && equal_or_match(node[:platform_version], version)
    end
  end
end

def violates_platform_constraint?(elements, role)
  if role_constraints[role] && role_constraints[role].has_key?("platform")
    matches_platform_constraint?(elements[role], role_constraints[role]["platform"])
  else
    false
  end
end

File: crowbar_framework/app/models/service_object.rb [hunk 17]
```diff
@@ -1576,4 +1616,26 @@ def profile(name, &block)
       block.call
     end
   end
+
+  def equal_or_match(value, maybe_regexp)
+    operator = lambda do
+      op, check = maybe_regexp.split(" ", 2)
+      value.to_f.send(op.to_sym, check.to_f)
```
vuntz: So, this is fancy, but actually, if we go the route of integrating operators for comparing versions, I think we should do it right :-)

The `to_f` bit is wrong, and we should instead split on `.` so we can handle versions like 12.04.1 (for Ubuntu), and do a proper comparison (12.04 < 12.04.1).

aplanas: I discussed this with @tboerger, but looks like that this is the expected behaviour for chef: '12.04.1'.to_f == 12.04, so '12.04' match with '== 12.04.1', but is not '< 12.04.1'.

But because we remove the '==' operator, (Chef do not provide it), if :platform_version is '12.04.1', it will not match with '12.04' restriction, and this is wrong IMHO.

To resolve that I will add a new check before the string comparison.

But I tend to disagree with the to_f approach for something that is not a number.

aplanas:
* Do not exit the loop (replace return, with return true if)

crowbar_framework/app/models/service_object.rb
vuntz: Well, I'm also worried about 12.04 vs 12.10 comparisons (which I think is something you raised earlier). Isn't it just broken right now?

crowbar_framework/spec/models/service_object_spec.rb
```diff
@@ -131,5 +131,43 @@
...
+          {
+            "dns-client" => {
+              "admin" => true ,
+              "platform" => { "suse" => "12.0" }
+            }
+          })
+        dns_service.validate_proposal_constraints(dns_proposal)
+        dns_service.validation_errors.first.should match(/can be used only for suse 12.0/)
+      end
+    end
```
vuntz: Thanks for adding tests. Feel free to ignore my comment here, but things I would add for better coverage:

* `exclude_platform` test
* test with more than one element in `platform` (for validating that the full hash is used)
* test with a proposal where nodes with different platforms are used

aplanas:
* Force to_f coercion to address some fancy versioning
* Add a new test case for Ubuntu numbering
* Extract invariant code from the loop

crowbar_framework/app/models/service_object.rb
aplanas: A new approach for this is in the JS and Ruby code of the amended commit.

aplanas:
* Do not coerce too much
* Add tests cases suggested by @vuntz in https://github.com/crowbar/barclamp-crowbar/pull/1175#discussion_r23442816

tboerger:
@aplanas Sorry, i have done it wrong... https://github.com/chef/chef/blob/master/lib/chef/version_class.rb should be a good starting point how to parse the versions correctly. Split the platform version my ```.``` and compare the array.

aplanas:
* Use match function only for version numbers (JS and Ruby)
* Implement a version comparator similar to https://github.com/chef/chef/blob/master/lib/chef/version_class.rb (JS and Ruby)

crowbar_framework/app/assets/javascripts/jquery/nodeList.js
```diff
@@ -359,6 +397,66 @@
...
+      }
+
+      match = /^(\d+)\.(\d+)$/.exec(version);
+      if (match) {
+        return [parseInt(match[1]), parseInt(match[2]), 0];
+      }
+
+      match = /^(\d+)$/.exec(version);
+      if (match) {
+        return [parseInt(match[1]), parseInt(match[2]), 0];
```
vuntz: s/parseInt(match[2])/0/

crowbar_framework/app/assets/javascripts/jquery/nodeList.js
```diff
@@ -359,6 +397,66 @@
...
+      return [0, 0, 0];
+    };
+
+    var cmp = function(op, version1, version2) {
+      var as_version1 = to_version(version1);
+      var as_version2 = to_version(version2);
+      var result = as_version1.map(function(v1, i) {
+        switch (op) {
+          case ">=":
+            return v1 >= as_version2[i];
```
vuntz: Shouldn't `v1` be `as_version1`?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1576,4 +1616,35 @@ def profile(name, &block)
...
+      end
+    end
+
+    cmp = lambda do |op, version|
+      (to_version.call(value).zip to_version.call(version)).all? do |v1, v2|
+        v1.send(op, v2)
+      end
+    end
+
+    op_value = maybe_regexp.scan(/^(>=?|<=?)\s*([.\d]+)$/)
```
vuntz: For the record, I like the code flow of the js bits a bit better (where this regexp is only used when we're not in the `/foo/` case.

crowbar_framework/app/assets/javascripts/jquery/nodeList.js
aplanas: v1 is the variable inside the map lambda

crowbar_framework/app/models/service_object.rb
tboerger: I think that's not a big issue as the regex is not too hungry.

aplanas: Can I maintain it here? I think that the ```eslif op_value``` is more clear here, and I can avoid multiple returns. I think that this is the idiomatic Ruby, but I am not sure.

aplanas:
* Handle properly single variant of version in JavaScript

crowbar_framework/app/assets/javascripts/jquery/nodeList.js
vuntz: Doh, read too quickly. Thanks!

crowbar_framework/app/models/service_object.rb
vuntz: @tboerger oh, it was not about performance, just about readability. If people prefer it here, then sure. It's just less readable (or well, more confusing) to me :-)

vuntz:
Let's do it!

PR #1176, opened 2015-01-15
Include form helper

bkutil:
The cinder backend UI uses a hack to workaround handlebars substituted
values and calls wrap_around. However, this is a method defined in form
helper. The helpers possibly might have been accessible from the DSL in rails 2,
but this is no longer the case in rails 4.

### DIFF
--- a/crowbar_framework/lib/dsl/proposal/attribute.rb
+++ b/crowbar_framework/lib/dsl/proposal/attribute.rb
@@ -204,7 +204,7 @@ module Dsl
       end
 
       def method_missing(method_name, *arguments, &block)
-        if view.respond_to? method_name
+        if view.respond_to?(method_name, true)
           view.send(method_name, *arguments, &block)
         else
           super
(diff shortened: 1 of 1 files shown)


tboerger:
Maybe this is related to the changed call of the dsl lib within the helper? I can not explain it otherwise...

bkutil:
@tboerger you're right. Maybe a wrong object gets passed as 'self' there.

bkutil:
Argh, scratch that. That seems correct. @tboerger What do you mean by 'changed call' then?

tboerger:
I think it is related to https://github.com/crowbar/barclamp-crowbar/commit/52edcd21746c4924a00acd23ea51d74a3b234be0#diff-034a177e23b728df7df90a05a91711c4L136
For ruby 2.1 i had to replace the block bind and call with ```instance_eval``` on block.

bkutil:
What seems to have changed is that the helper methods are now protected. respond_to? then returns false unless explicitly told to search protected & private methods. Will submit a fixed patch in a moment.

tboerger:
Sounds great, i have not tested it but you get my :+1:

vuntz:
No idea if it's correct, but that fixes the issue for me.

Since @tboerger gave a +1, let's merge.

PR #1177, opened 2015-01-16
Install uglifier, not closure-compiler

dirkmueller:
We use sprockets with uglifier, there is no point in installing
closure-compiler


tboerger:
Yes, that's a wrong dependency :+1:

PR #1178, opened 2015-01-16
Use soft reboot/shutdown by default and provide hard reboot/shutdown actions

vuntz:
Note that the hard reboot/shutdown methods are not exposed in the UI.

### DIFF
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -1212,6 +1212,16 @@ class NodeObject < ChefObject
     bmc_cmd("power on")
   end
 
+  def powercycle
+    set_state("reboot")
+    bmc_cmd("power cycle")
+  end
+
+  def poweroff
+    set_state("shutdown")
+    bmc_cmd("power off")
+  end
+
   def identify
     bmc_cmd("chassis identify")
   end
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -1089,6 +1089,20 @@ class NodeObject < ChefObject
     @node["ipmi"]["bmc_password"] rescue nil
   end
 
+  def ssh_cmd(cmd)
+    # Have to redirect stdin, stdout, stderr and background reboot
+    # command on the client else ssh never disconnects when client dies
+    # `timeout` and '-o ConnectTimeout=10' are there in case anything
+    # else goes wrong...
+    unless system("sudo", "-i", "-u", "root", "--",
+                  "timeout", "-k", "5s", "15s",
+                  "ssh", "-o", "ConnectTimeout=10", "root@#{@node.name}",
+                  "#{cmd} </dev/null >/dev/null 2>&1 &")
+      Rails.logger.warn("ssh command \"#{cmd}\" for #{@node.name} failed - node in unknown state")
+      return nil
+    end
+  end
+
   def bmc_cmd(cmd)
     if bmc_address.nil? || get_bmc_user.nil? || get_bmc_password.nil? ||
         !system("ipmitool", "-I", "lanplus", "-H", bmc_address, "-U", get_bmc_user, "-P", get_bmc_password, cmd)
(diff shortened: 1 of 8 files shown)


jsuchome:
Nice. But yes, we should have it in UI (probably in some other request).

I do not really like the name, 'hard reboot'. Wouldn't 'force reboot' be better? Or maybe it is usually called this way and my language feeling is wrong.

vuntz:
@jsuchome http://en.wikipedia.org/wiki/Reboot_%28computing%29 :-)

jsuchome:
Not that it really matters, but that article defines cold/hard reboot as "when power to the system is physically turned off and back on again". Which is not what ``reboot -f`` does.

vuntz:
Btw, the reason I haven't put that in the UI is that I don't know how to add the buttons without having it look ugly because of too many buttons.

vuntz:
> Not that it really matters, but that article defines cold/hard reboot as "when power to the system is physically turned off and back on again". Which is not what ``reboot -f`` does.

Well, when IPMI is used, that's what it does. But you're right, I guess. I'm happy to rename to force reboot/shutdown if others agree it's better.

jsuchome:
Yes, it probably needs a touch from UI guys :-)

tboerger:
Maybe we can add a button group with a drodown list for all these actions. That way it stays intact even for responsive small screen widths.

vuntz:
@tboerger big +1. No idea how to do that, though :-)

tboerger:
```html
<div class="dropdown">
  <button class="btn btn-default dropdown-toggle" type="button" data-toggle="dropdown" aria-expanded="true">
    Actions
    <span class="caret"></span>
  </button>
  <ul class="dropdown-menu" role="menu">
    <li role="presentation"><a role="menuitem" href="#">Action</a></li>
    <li role="presentation"><a role="menuitem" href="#">Another action</a></li>
    <li role="presentation"><a role="menuitem" href="#">Something else here</a></li>
    <li role="presentation"><a role="menuitem" href="#">Separated link</a></li>
  </ul>
</div>
```

jsuchome:
@tboerger could  you reformulate it to haml? Currently we have https://github.com/crowbar/barclamp-crowbar/blob/master/crowbar_framework/app/views/nodes/_buttons_power.html.haml

tboerger:
```ruby
.dropdown
  %button.btn.btn-default.dropdown-toggle{ type: "button", data: { toggle: "dropdown" } }
    = t(".actions")
    %span.caret

  %ul.dropdown-menu{ role: "menu" }
    - if @node.admin?
      - if @node.bmc_set?
        %li{ role: "presentation" }
          = link_to t(".identify"), hit_node_path(@node.handle, "identify"), role: "menuitem", remote: true
    - else
      %li{ role: "presentation" }
        = link_to t(".reboot"), hit_node_path(@node.handle, "reboot"), role: "menuitem", remote: true
      %li{ role: "presentation" }
        = link_to t(".hardreboot"), hit_node_path(@node.handle, "hardreboot"), role: "menuitem", remote: true
      %li{ role: "presentation" }
        = link_to t(".shutdown"), hit_node_path(@node.handle, "shutdown"), role: "menuitem", remote: true
      %li{ role: "presentation" }
        = link_to t(".hardshutdown"), hit_node_path(@node.handle, "hardshutdown"), role: "menuitem", remote: true

      - if @node.bmc_set?
        %li{ role: "presentation" }
          = link_to t(".poweron"), hit_node_path(@node.handle, "poweron"), role: "menuitem", remote: true
        %li{ role: "presentation" }
          = link_to t(".identify"), hit_node_path(@node.handle, "identify"), role: "menuitem", remote: true
```

I have not tested it, so please check if the ```data-remote``` attributes are really set and if it looks correct. Don't miss the i18n keys :).

crowbar_framework/app/models/node_object.rb
```diff
@@ -1169,7 +1176,7 @@ def set_state(state)
         Rails.logger.warn("chef client seems to be still running after 5 minutes of wait; going on with the reboot")
       end
 
-      bmc_cmd("power cycle")
+      ssh_cmd("/sbin/reboot")
```
dirkmueller: that doesn't work on windows :-)

tboerger: Correct, just call shitdown -r without a path.

vuntz:
I renamed the hard reboot/hard shutdown bits to power cycle/power off, as it works nicely with the "power on" bit.

aplanas:
LGTM if is expected to work only on Linux (calls to /sbin/* commands and references to /var/run/crowbar/chef-client.* files)

Maybe we can add a check to disable this command in Windows (node[:platform] == 'windows') in _buttons_power.html.haml

jsuchome:
+1

crowbar_framework/app/models/node_object.rb
tboerger: What about this? If we call it only ```shutdown -r``` it should work on windows and linux platform.

vuntz: There's no ssh server on the hyperv nodes, so no, it won't work.

tboerger: So we should hide the buttons for windows machines.

vuntz: Yes, but it's unrelated to this PR. Doing another PR for it right now :-)

vuntz: Done in https://github.com/crowbar/barclamp-crowbar/pull/1202

PR #1180, opened 2015-01-19
make logging more consistent

MaximilianMeister:
right now logging is a bit inconsistent. 
see for example `AR:` vs `apply_role` as a prefix for the logging output of `service_object.rb`
some time ago i even had to ask what `AR` means, it was not so clear.
one way to generalize it is rubys magic `__method__` which just gives you the name of the method, where the statement is executed.
this would have to be done for the other barclamps as well.

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -30,18 +30,18 @@ class CrowbarService < ServiceObject
     return [404, "No state specified"] if state.nil?
     # FIXME: validate state
 
-    @logger.info("Crowbar transition enter: #{name} to #{state}")
+    @logger.info("#{__method__} enter: #{name} to #{state}")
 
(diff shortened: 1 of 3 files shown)


aspiers:
This is a really nice idea but I think it should really be done by a helper.  That way we can change the format everywhere just by changing one place.  Of course this will make it a bit more difficult since `__method__` cannot be used directly.  But it is still possible to do.

aspiers:
To clarify, I'm thinking of an API something like:

    @logger.mdebug "foo"

where `m` stands for "method".

tboerger:
I don't like this approach. I would prefer the already intergrated tagged loghing functionality: http://api.rubyonrails.org/classes/ActiveSupport/TaggedLogging.html

MaximilianMeister:
@aspiers @tboerger agreed.
closing this, as there are better ways to do this

PR #1181, opened 2015-01-19
Disable root element in API JSONs (bnc#912309)

bkutil:
This supersedes https://github.com/crowbar/barclamp-crowbar/pull/1171

Configure the ActiveModel::Serializers gem to not use root element in
the rendered JSON responses.

This keeps our API JSON responses consistent between cloud releases.

### DIFF
--- a/crowbar_framework/config/initializers/serializer.rb
+++ b/crowbar_framework/config/initializers/serializer.rb
@@ -1,3 +1,19 @@
+#
+# Copyright 2014, SUSE LINUX Products GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


dirkmueller:
"LGTM"

tboerger:
Is there no config option like https://github.com/crowbar/barclamp-crowbar/blob/master/crowbar_framework/config/application.rb#L30? And please add the copyright stuff as on all source files.

tboerger:
What's wrong with ```default_serializer_options``` within ApplicationController? That way we can change it to a default behavior with something like API version 2.

bkutil:
> Is there no config option like...

I checked the [readme](https://github.com/rails-api/active_model_serializers/blob/0-9-stable/README.md) and there does not seem to be one.

> What's wrong with default_serializer_options within ApplicationController

I think of this rather as of a global setting, which should be set once when the app starts, not checked on every request. I'm fine to change that to a callback in app controller when we have api v2 :)

tboerger:
I have not tried it on my own, but may does this work?

```ruby
ActiveModel::Serializer.setup do |config|
  config.root = false
end
```

Looks IMHO a little bit better :)

bkutil:
Agreed, unfortunately, it does not work :<

tboerger:
ok so :+1:

vuntz:
Tested, +1

PR #1182, opened 2015-01-19
Update rspec rails to 3.1 compatible with rails 4

bkutil:
This prevents

`named_routes.helpers` is deprecated, please use `route_defined?

deprecation warningis.

### DIFF
--- a/crowbar_framework/Gemfile
+++ b/crowbar_framework/Gemfile
@@ -58,7 +58,7 @@ gem "rack-mini-profiler", "~> 0.9",
 
 group :development, :test do
   gem "brakeman", "~> 2.6"
-  gem "rspec-rails", "~> 2.14"
+  gem "rspec-rails", "~> 3.1"
 end
 
 group :test do


vuntz:
Can't comment on the change itself, but typo in the commit message: warningis

bkutil:
@vuntz my Engrish is so powderful! Thanks :)

tboerger:
The rpm's have to be updated as well please. :+1:

bkutil:
OK, in that case, please wait with the merge. I'll prepare the packages first.

crowbar_framework/Gemfile
```diff
@@ -58,7 +58,7 @@ gem "rack-mini-profiler", "~> 0.9",
 
 group :development, :test do
   gem "brakeman", "~> 2.6"
-  gem "rspec-rails", "~> 2.14"
+  gem "rspec-rails", "~> 3.1"
```
dirkmueller: this should be '~> 3.1.0' to prevent picking 3.2.

tboerger: No, all gems are set to ~> x.x as we always reuse the gems installed through rpm.

dirkmueller:
@bkutil package is updated already, was trivial

crowbar_framework/Gemfile
dirkmueller: I'm not against reusing the gems through rpm at all (the rpm packaged version is not shipping Gemfile anyway so your point ins mood) but I would like to ensure that the dev environment is reasonably similar to the production environment to avoid accidental breakages from slipping in.

bkutil:
@dirkmueller cool, thanks.

crowbar_framework/Gemfile
tboerger: Why do you care about the versions in the Gemfile? You are not using it anyway. And if we pin the versions too hard we get problems for updates as we always have to ship a new crowbar version instead of just updating the gem rpm. I have not done this version scheme just out of fun...

tboerger:
> @dirkmueller cool, thanks. Can we merge, then?

Not all required rubygems got an update. The rspec dependencies are still missing.

crowbar_framework/Gemfile
dirkmueller: I would like to have the travis / developer environment be reasonably similar to the production environment to make sure that there are no uncaught accidental breakages.

tboerger: It is similar. and i have tested all the version constraints over weeks pretty hard.

dirkmueller: It is not, it installs e.g. Rails 4.2 while we have Rails 4.1. There are many such changes.

dirkmueller:
we really need this change to unbreak travis, so merging..

PR #1183, opened 2015-01-19
Rename serve_static_assets to serve_static_files

bkutil:
Rename the serve_static_assets to serve_static_files in the environment
configuration. This prevents

The configuration option `config.serve_static_assets` has been renamed

deprecation warnings (the option will be removed in rails 5).


tboerger:
:+1:

PR #1184, opened 2015-01-19
Followup on respond_to fix for DSL (PR#1176)

bkutil:
I realized that my fix for https://github.com/crowbar/barclamp-crowbar/pull/1176 was incomplete - in fact, the same issue can happen in deployment.rb as well. This patch fixes that.


vuntz:
+1

tboerger:
:+1:

PR #1185, opened 2015-01-19
Ruby 2.1 fixes for crowbar batch

vuntz:
These are changes from https://github.com/crowbar/barclamp-crowbar/pull/1171 that would still be valid after https://github.com/crowbar/barclamp-crowbar/pull/1181 is merged.


bkutil:
+1

tboerger:
:+1:

PR #1191, opened 2015-01-20
Go to dashboard on deleted node edit (bnc#913818)

bkutil:
If a node is forgotten and then edited before it disappears, we get an
exception.

As the edit action is purely navigational (i.e. not called from an api),
redirect to the dashboard and displaying a flash notification should be
sufficient.

### DIFF
--- a/crowbar_framework/app/controllers/nodes_controller.rb
+++ b/crowbar_framework/app/controllers/nodes_controller.rb
@@ -313,7 +313,7 @@ class NodesController < ApplicationController
   def edit
     get_node_and_network(params[:id] || params[:name])
     if @node.nil?
-      flash[:notice] = "Node #{params[:id] || params[:name]} not found."
+      flash[:alert] = "Node #{params[:id] || params[:name]} not found."
       return redirect_to nodes_path
     end
   end
--- a/crowbar_framework/app/controllers/nodes_controller.rb
+++ b/crowbar_framework/app/controllers/nodes_controller.rb
@@ -312,6 +312,10 @@ class NodesController < ApplicationController
 
   def edit
     get_node_and_network(params[:id] || params[:name])
+    if @node.nil?
+      flash[:notice] = "Node #{params[:id] || params[:name]}: not found"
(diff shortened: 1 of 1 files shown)


tboerger:
I would use an exception within the get method and fetch it with a better error message with rescue_from.

or is you dont like to use an exception do it anyway within the get method.

bkutil:
@tboerger that sounds reasonable. I'm not sure though - should all actions fail if @node is nil, or are there any for which nil is ok?

vmoravec:
The patch works fine. Btw the colon following the node name looks a bit strange: `Node d52-54-03-77-77-03: not found` , it looks more natural without it.

vmoravec:
+1

bkutil:
@tboerger I added your suggestion to trello, so it won't get lost.

tboerger:
@bkutil great

tboerger:
> @tboerger that sounds reasonable. I'm not sure though - should all actions fail if node is nil, or are there any for which nil is ok?

IMHO all of the should fail because they doesnt make sense if the node object is not retrievable correctlty.

crowbar_framework/app/controllers/nodes_controller.rb
```diff
@@ -312,6 +312,10 @@ def show
 
   def edit
     get_node_and_network(params[:id] || params[:name])
+    if @node.nil?
+      flash[:notice] = "Node #{params[:id] || params[:name]} not found."
```
vuntz: Shouldn't this be `flash[:alert]`? (not sure)

vuntz:
Just one question, but +1 in general

crowbar_framework/app/controllers/nodes_controller.rb
bkutil: Yeah, I think it might be better to have this as an alert.

tboerger: :notice will be green and :alert will be red :)

PR #1192, opened 2015-01-20
Fix logging format (bnc#912805)

bkutil:
> The config.log_formatter setting does not have any effect on the
> actual log format, if TaggedLogger or AS::Logger is passed via config.logger.
> 
> This might be due to the way the TaggedLogger is initialized [1]. To
> make it work, we would have to first instantiate the logger, then set
> its formatter, then pass it to the tagged logging. This is ugly.
>
> We could also use ActiveSupport::Logger instead, but the
> config.log_formatter does not work there, either (unless we again set
> the formatter as an attribute).

As Thomas correctly pointed out, the config.log_format does not apply if the config.logger is set.

Let's just use simple logger for now, and add tagged logging later. Imho there is no fast and easy way how to keep both tagged logging and old log format (unless we implement a custom formatter that does both).

### DIFF
--- a/crowbar_framework/config/environments/development.rb
+++ b/crowbar_framework/config/environments/development.rb
@@ -52,8 +52,7 @@ Rails.application.configure do
   config.log_tags = []
 
   config.logger = ActiveSupport::TaggedLogging.new(
-    Logger.new Rails.root.join("log", "development.log") # SyslogLogger.new
+    Logger.new(Rails.root.join("log", "development.log"))
   )
-
-  config.log_formatter = ::Logger::Formatter.new
+  config.logger.formatter = ::Crowbar::Logger::Formatter.new
 end
--- a/crowbar_framework/config/environments/production.rb
+++ b/crowbar_framework/config/environments/production.rb
@@ -52,8 +52,7 @@ Rails.application.configure do
   config.log_tags = []
 
   config.logger = ActiveSupport::TaggedLogging.new(
-    Logger.new File.join(ENV["CROWBAR_LOG_DIR"], "production.log") # SyslogLogger.new
+    Logger.new(File.join(ENV["CROWBAR_LOG_DIR"], "production.log"))
   )
-
-  config.log_formatter = ::Logger::Formatter.new
+  config.logger.formatter = ::Crowbar::Logger::Formatter.new
 end
--- a/crowbar_framework/config/environments/test.rb
+++ b/crowbar_framework/config/environments/test.rb
@@ -52,8 +52,7 @@ Rails.application.configure do
   config.log_tags = []
 
   config.logger = ActiveSupport::TaggedLogging.new(
-    Logger.new Rails.root.join("log", "test.log") # SyslogLogger.new
+    Logger.new(Rails.root.join("log", "test.log"))
   )
-
-  config.log_formatter = ::Logger::Formatter.new
+  config.logger.formatter = ::Crowbar::Logger::Formatter.new
 end
--- a/crowbar_framework/lib/crowbar/logger/formatter.rb
+++ b/crowbar_framework/lib/crowbar/logger/formatter.rb
@@ -1,0 +1,38 @@
+#
+# Copyright 2014, SUSE LINUX Products GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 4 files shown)


tboerger:
I really want to keep the tagged logging as it makes sense on so many places of crowbar. So maybe let's take the example from https://cbpowell.wordpress.com/2013/08/09/beautiful-logging-for-ruby-on-rails-4/ and overwrite ```ActiveSupport::Logger::SimpleFormatter```.

tboerger:
Just for the notes: As you can see at https://github.com/rails/rails/blob/master/railties/lib/rails/application/bootstrap.rb#L32-L55 ```log_formatter``` config value gets only used if ```logger``` is not set.

bkutil:
> Just for the notes: As you can see at https://github.com/rails/rails/blob/master/railties/lib/rails/application/bootstrap.rb#L32-L55 ```log_formatter``` config value gets only used if ```logger``` is not set.

Yeah. You're right.

crowbar_framework/config/environments/development.rb
```diff
@@ -51,9 +51,7 @@
   config.log_level = :debug
   config.log_tags = []
 
-  config.logger = ActiveSupport::TaggedLogging.new(
-    Logger.new Rails.root.join("log", "development.log") # SyslogLogger.new
-  )
-
-  config.log_formatter = ::Logger::Formatter.new
+  logger = ActiveSupport::Logger.new(Rails.root.join("log", "development.log"))
+  logger.formatter = ::Logger::Formatter.new
```
tboerger: ```Logger::Formatter``` won't work as it doesnt respect the tags. I think we have to create a formatter within lib folder and set it here.

bkutil:
@tboerger I kept the tagged logger.

crowbar_framework/config/environments/production.rb
```diff
@@ -51,9 +51,7 @@
   config.log_level = :debug
   config.log_tags = []
 
-  config.logger = ActiveSupport::TaggedLogging.new(
-    Logger.new File.join(ENV["CROWBAR_LOG_DIR"], "production.log") # SyslogLogger.new
-  )
-
-  config.log_formatter = ::Logger::Formatter.new
+  logger = ActiveSupport::Logger.new(File.join(ENV["CROWBAR_LOG_DIR"], "production.log"))
+  logger.formatter = ::Logger::Formatter.new
```
tboerger: Same as above

crowbar_framework/config/environments/test.rb
```diff
@@ -51,9 +51,7 @@
   config.log_level = :debug
   config.log_tags = []
 
-  config.logger = ActiveSupport::TaggedLogging.new(
-    Logger.new Rails.root.join("log", "test.log") # SyslogLogger.new
-  )
-
-  config.log_formatter = ::Logger::Formatter.new
+  logger = ActiveSupport::Logger.new(Rails.root.join("log", "test.log"))
+  logger.formatter = ::Logger::Formatter.new
```
tboerger: Same as above

crowbar_framework/config/environments/development.rb
bkutil: OK. I won't be able to do that before the release. What do you suggest instead?

crowbar_framework/config/environments/production.rb
bkutil: Same as above :)

crowbar_framework/config/environments/test.rb
bkutil: Same as above :)

tboerger:
Untested but give this a try:

```ruby
logger.formatter = ::Crowbar::Logger::Formatter.new
```

```ruby
module Crowbar
  module Logger
    class Formatter < ::Logger::Formatter
      include ActiveSupport::TaggedLogging::Formatter

      def call(severity, timestamp, progname, msg)
        final_message = if msg.is_a? String
          msg
        else
          msg.inspect
        end

        super(
          severity,
          timestamp,
          progname,
          final_message
        )
      end
    end
  end
end
```

:)

tboerger:
I have added this to the production environment:

```ruby
config.logger = ActiveSupport::TaggedLogging.new(
  Logger.new Rails.root.join(ENV["CROWBAR_LOG_DIR"], "production.log") # SyslogLogger.new
)

config.logger.formatter = ::Crowbar::Logger::Formatter.new
```

And the above class to ```crowbar_framework/lib/crowbar/logger/formatter.rb```

vuntz:
+1

tboerger:
:+1:

bkutil:
@tboerger thanks for looking into this and providing the actual fix :)

tboerger:
@bkutil whenever you want again :)

PR #1194, opened 2015-01-21
Fix various dependencies to be closer to production

dirkmueller:
To ensure that development and production environments are
somewhat compareable os that the added rspec tests have an
actual meaning, the dependencies are locked on the 3rd level
of the semantic versioning.

### DIFF
--- a/crowbar_framework/Gemfile
+++ b/crowbar_framework/Gemfile
@@ -17,54 +17,53 @@
 
 source "https://rubygems.org"
 
-gem "rails", "~> 4.1"
-gem "haml-rails", "~> 0.5"
-gem "sass-rails", "~> 4.0"
-gem "rainbows-rails", "~> 1.0"
+gem "rails", "~> 4.1.9"
+gem "haml-rails", "~> 0.5.3"
+gem "sass-rails", "~> 4.0.5"
+gem "rainbows-rails", "~> 1.0.1"
+gem "active_model_serializers", "~> 0.9.0"
+gem "activeresource", "~> 4.0.0"
+gem "closure-compiler", "~> 1.1.10"
+gem "dotenv", "~> 1.0.2"
+gem "hashie", "~> 3.3.1"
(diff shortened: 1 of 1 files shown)


tboerger:
:-1: i prefer to add the Gemfile.lock just for depevelopment dependencies. For packaging the less restrictive constrai ts are much better.

tboerger:
Needs a rebase and do that for all gems please.

vuntz:
+1.
@tboerger what do you think?

tboerger:
Maybe do it directly for all? Including development and test?

rsalevsky:
Whats the current status here?

aspiers:
I'm not sure I have enough context to vote, but I think probably :+1: from me, since in *theory* `~> x.y.z` should avoid breakage for any gem which follows SemVer.  Of course plenty of gems don't, but the alternative is to manually check each gem to see whether it does, and then adjust `Gemfile` accordingly ... good luck with that.

mjura:
+1

tboerger:
Still -0.5 because it should be done for all and not only partially.

rsalevsky:
+1

crowbar_framework/Gemfile
```diff
@@ -17,52 +17,51 @@
...
 
 group :test do
-  gem "mocha", "~> 1.1"
-  gem "sinatra", "~> 1.4"
-  gem "webmock", "~> 1.19"
+  gem "mocha", "~> 1.1.0"
+  gem "sinatra", "~> 1.4.5"
+  gem "webmock", "~> 1.19.0"
 
   gem "simplecov", "~> 0.9", require: false
```
tboerger: You missed this one

tboerger:
Just ```simplecov```, afterwards we can merge it.

rsalevsky:
Should be `"simplecov", "~> 0.9.1"`

dirkmueller:
@rsalevsky done

PR #1196, opened 2015-01-21
Exclude SLE12 platform.

aplanas:
Depends on crowbar/barclamp-crowbar#1175

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -18,6 +18,21 @@
 class CrowbarService < ServiceObject
   attr_accessor :transition_save_node
 
+  class << self
+    def role_constraints
+      {
+        "crowbar" => {
+          "count" => 1,
+          "admin" => true,
+          "exclude_platform" => {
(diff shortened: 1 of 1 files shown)


tboerger:
Do we really want to exclude that hard?

aplanas:
@tboerger We are using this as a reference to decide what to exclude: https://etherpad.nue.suse.com/p/roles

jsuchome:
+1

tboerger:
Does this constraint really make sense here? You can not drag and drop any role anyway here.

aplanas:
@tboerger This is also my opinion, based on my limited knowledge, but I found it useful for documentation.

So, is safe to remove the constraint here?

vuntz:
I would actually leave it for consistency. But it's true that it doesn't matter much.

aplanas:
* Exclude Windows platforms

aplanas:
* Add "count" and "admin" constraints for completeness.

vuntz:
+1

dirkmueller:
+1

PR #1198, opened 2015-01-21
Add CrowbarHelper.in_sledgehammer? helper

vuntz:
We have this code in a few barclamps, and it's not really understandable
at first. So the small helper will make it clearer what it's for, and
avoid copy & paste bugs.


tboerger:
LGTM

PR #1199, opened 2015-01-21
Only run travis against Ruby 2.1

dirkmueller:
Older versions won't work anymore since we broke compatibility.


dirkmueller:
merging since its obvious.

PR #1200, opened 2015-01-23
Implement a proper comparison for versions.

aplanas:
In crowbar/barclamp-crowbar#1175 I introduced a wrong comparison
algorithm after missunderstanding how <=> works on Ruby.

This commit fix the issue and 8+ test cases to cover all operators.

### DIFF
--- a/crowbar_framework/app/assets/javascripts/jquery/nodeList.js
+++ b/crowbar_framework/app/assets/javascripts/jquery/nodeList.js
@@ -417,29 +417,40 @@
       return [0, 0, 0];
     };
 
-    var cmp = function(op, version1, version2) {
-      var as_version1 = to_version(version1);
-      var as_version2 = to_version(version2);
-      var result = as_version1.map(function(v1, i) {
-        switch (op) {
-          case ">=":
-            return v1 >= as_version2[i];
-          case ">":
-            return v1 > as_version2[i];
-          case "<=":
-            return v1 <= as_version2[i];
-          case "<":
-            return v1 < as_version2[i];
-          case "==":
-            return v1 == as_version2[i];
-          default:
-            return false;
+    var cmp = function(version1, version2) {
+      var ans = 0;
+      // version1 and version2 are arrays of the same length.  At this
+      // point this assertion is true:
+      // assert(version1.length == version2.length)
+      for (var i=0; i<version1.length; i++) {
+        if (version1[i] < version2[i]) {
+          ans = -1;
+        } else if (version1[i] > version2[i]) {
+          ans = 1;
         }
-      });
+        if (ans != 0) {
+          return ans;
+        }
+      }
+      return 0;
+    };
 
-      return result.every(function(element, index, array) {
-        return element;
-      });
+    var oper = function(op, version1, version2) {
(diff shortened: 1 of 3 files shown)


aplanas:
review? @tboerger @bkutil @vuntz

tboerger:
Review at friday evening? :)
Looks good for me, but i have not tested it.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1631,9 +1631,29 @@ def equal_or_match_version(value, maybe_regexp)
       end
     end
 
-    cmp = lambda do |op, version|
-      (to_version.call(value).zip to_version.call(version)).all? do |v1, v2|
-        v1.send(op, v2)
+    cmp = lambda do |version1, version2|
+      version1.zip version2.each do |v|
```
vuntz: I find this line hard to parse without parenthesis. Is this the same as `version1.zip(version2).each`? If yes, could you rewrite it like that?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1642,9 +1662,9 @@ def equal_or_match_version(value, maybe_regexp)
     if maybe_regexp.start_with?('/') && maybe_regexp.end_with?('/')
       Regexp.new(maybe_regexp[1..-2]).match(value)
     elsif op_value.length > 0
-      cmp.call(op_value.first.first, op_value.first.last)
+      oper.call(op_value.first.first, op_value.first.last)
```
vuntz: For the record, reading this line is actually hard without reparsing the previous lines. Might be good to use some (arguably useless) variables just to get some good variable names for `op_value.first`, or `op_value.first.first`, etc.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1631,9 +1631,29 @@ def equal_or_match_version(value, maybe_regexp)
...
-        v1.send(op, v2)
+    cmp = lambda do |version1, version2|
+      version1.zip version2.each do |v|
+        ans = v[0] <=> v[1]
+        return ans if ans != 0
+      end
+      0
+    end
+
+    oper = lambda do |op, version|
```
vuntz: Again, just for readability: I'd make ` value` an argument of this, so we don't have to scroll back to see that it's an argument to the parent method.

crowbar_framework/app/assets/javascripts/jquery/nodeList.js
```diff
@@ -417,29 +417,37 @@
...
-          case "<":
-            return v1 < as_version2[i];
-          case "==":
-            return v1 == as_version2[i];
-          default:
-            return false;
+    var cmp = function(version1, version2) {
+      var ans = 0;
+      for (var i=0; i<version1.length; i++) {
+        if (version1[i] < version2[i]) {
```
vuntz: There's a (correct, iirc) assumption to at least document, or to remove from the code, that `version1.length == version2.length`.

aplanas:
* Improve code readability

aplanas:
* Add comment in nodeList.js to explain assertion

vuntz:
+1

PR #1201, opened 2015-01-26
Use new platform constraints for element_info method

vuntz:
This method is used to find out which node can be used for a specific
role, so using the platform constraints there makes sense.


tboerger:
LGTM

PR #1202, opened 2015-01-26
Hide power actions UI for windows nodes

vuntz:
All actions that required ssh can't work for windows nodes since there's
no ssh server installed there.

Also add warning when trying to use API requiring ssh for windows nodes


aplanas:
+1

jsuchome:
+1

tboerger:
Why not :+1:

PR #1203, opened 2015-01-28
Removed ember-js dependency

tboerger:
As we are currently not using ember-js i have removed the dependency for
this release that we don't need to ship dependencies like nodejs.


dirkmueller:
+1

PR #1205, opened 2015-01-28
No need for specific gem version on travis anymore

tboerger:


tboerger:
@bkutil Do you have any idea why it fails? https://rubygems.org/gems/chef/versions shows that there are far more newer versions than chef 10.26, and the version constraint is set to ```~> 10.24```. And for the gem version, we are running ```2.2.2``` and travis installs ```2.2.0```.

tboerger:
So, https://github.com/chef/chef/blame/10-stable/chef/lib/chef/provider/package/rubygems.rb#L37-L41 this have been changed a few days after release of ```10.26```, so we need at least a more current version of chef. Why is it installing ```10.26``` instead of ```10.34```?

tboerger:
So travis succeeds now. Merge? :)

bkutil:
This patch is good, although it might be better to wait w/ the merge after release - just so that some change between 10.26 and 10.34 does not cause regression. What do you think?

tboerger:
I think merging this should be no issue as we are anyway running on a more current version. I will increase the version exactly to our version.

vuntz:
+1 (because this matches the version we're using)

bkutil:
I see, fine w/ me then.

PR #1206, opened 2015-01-28
Update package installation list here as well

dirkmueller:
I'm not sure there is really a lot of use for
this right now here since crowbar-barclamp-crowbar should have
the right dependencies already, but I don't want to play
with that right now.

### DIFF
--- a/chef/cookbooks/crowbar/recipes/default.rb
+++ b/chef/cookbooks/crowbar/recipes/default.rb
@@ -56,6 +56,7 @@ when "suse"
 
     ruby2.1-rubygem-activerecord-session_store
     ruby2.1-rubygem-activeresource
+    ruby2.1-rubygem-active_model_serializers
     ruby2.1-rubygem-chef
     ruby2.1-rubygem-closure-compiler
     ruby2.1-rubygem-dotenv


tboerger:
:-1: uglifier is still used.

dirkmueller:
during runtime? but uglifier depends on execjs which depends on nodejs.. I thought thats what you're trying to change with the precompile step ?

tboerger:
Currently it's still part of the Gemfile and boot.rb, the problem with the runtime dependency is not resolved for barclamp-crowbar for now.

dirkmueller:
hmm. okay.

vuntz:
@tboerger can you revisit this one and review it again?

tboerger:
@vuntz now it makes sense to add rubygem-closure-compiler instead of rubygem-uglifier-2_2.

dirkmueller:
@tboerger closure-compiler was already in the list. Resolved merge conflict (it seems somebody else was able to get in the same patch than I originally proposed)

tboerger:
@dirkmueller yeah, it has been part of tje asset precompile pull request.

tboerger:
I habe not checked if the list got synced full against the Gemfile, but as long as it is in sync i give my +1

tboerger:
@vuntz you merged too fast. I ahve checked the gems now and we are misssing ```i18n-js``` and ```rack-mini-profiler```.

PR #1207, opened 2015-01-28
States validation

bkutil:
https://bugzilla.suse.com/show_bug.cgi?id=907440

### DIFF
--- a/crowbar_framework/app/controllers/barclamp_controller.rb
+++ b/crowbar_framework/app/controllers/barclamp_controller.rb
@@ -67,17 +67,20 @@ class BarclampController < ApplicationController
     state = params[:state] # State of node transitioning
     name = params[:name] # Name of node transitioning
 
-    status, response = @service_object.transition(id, name, state)
-
-    if status != 200
-      render :text => response, :status => status
+    unless NodeObject::API_REACHABLE_STATES.include?(state)
+      render :text => "State #{state} is invalid", :status => 400
     else
-      # Be backward compatible with barclamps returning a node hash, passing
-      # them intact.
-      if response[:name]
-        render :json => NodeObject.find_node_by_name(response[:name]).to_hash
+      status, response = @service_object.transition(id, name, state)
+      if status != 200
+        render :text => response, :status => status
       else
-        render :json => response
+        # Be backward compatible with barclamps returning a node hash, passing
+        # them intact.
+        if response[:name]
+          render :json => NodeObject.find_node_by_name(response[:name]).to_hash
+        else
+          render :json => response
+        end
       end
     end
   end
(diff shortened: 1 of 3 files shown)


crowbar_framework/app/models/node_object.rb
```diff
@@ -19,6 +19,13 @@
 require 'timeout'
 
 class NodeObject < ChefObject
+  API_REACHABLE_STATES = [
+    "applying", "discovered", "discovering", "hardware-install",
```
vuntz: Sorry,  I keep discovering new things in this list :-) Can we remove `"hardware-install"`?

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -67,17 +67,20 @@ def transition
     state = params[:state] # State of node transitioning
     name = params[:name] # Name of node transitioning
 
-    status, response = @service_object.transition(id, name, state)
-
-    if status != 200
-      render :text => response, :status => status
+    unless NodeObject::API_REACHABLE_STATES.include?(state)
+      render :text => "State #{state} is invalid", :status => 400
```
dirkmueller: perhaps some quoting 

"State '#{state}' is not a valid state"

just in case there are some strange whitespace issues..

tboerger:
IMHO this validation have to be done within the model and not within the controller.

tboerger:
And i would like to see some kind of raw transition where you can transition into any state while development.

bkutil:
@tboerger which model would you suggest? Node object does not have any code related to transitions transitioning code at all and service object defines an 'abstract interface' for other barclamp services, so I'm not sure if adding another public method makes sense there.

tboerger:
The validation should be done within the ```transition``` method, just return ```[400, "State #{state} is invalid"]``` from that method if the state is not allowed.

bkutil:
Won't that fail if any barclamp overrides transition and does not call super?

tboerger:
@bkutil I don't know how many barclamps are doing that. Another good reason to integrate hooks into the models like ```before_save``` on active model :P

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -67,17 +67,20 @@ def transition
     state = params[:state] # State of node transitioning
     name = params[:name] # Name of node transitioning
 
-    status, response = @service_object.transition(id, name, state)
-
-    if status != 200
-      render :text => response, :status => status
+    unless valid_transition_states.include?(state)
+      render :text => "State '#{state}' is not valid.", :status => 400
```
vuntz: Just wondering if `400` is the best error code, but other than that, +1

tboerger:
I think 400 is ok for that. And as the commit messages say it's currently not really integrateable into the model so :+1:

tboerger:
I have added an entry to our idea collection for crowbar.

PR #1208, opened 2015-01-29
Add select_nodes_for_role helper to help create default proposals

vuntz:
This will select nodes for roles, based on constraints.

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -592,11 +592,7 @@ class ServiceObject
     # from the nodes list in the case the new proposal would not be valid, so
     # nodes that can't be added at all would not be returned.
     nodes.reject! do |node|
-      elements = { role.to_s => [node] }
-      violates_admin_constraint?(elements, role) ||
-        violates_platform_constraint?(elements, role) ||
-        violates_exclude_platform_constraint?(elements, role) ||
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/service_object.rb
```diff
@@ -649,6 +645,33 @@ def find_dep_proposal(bc, optional=false)
...
+    violates_admin_constraint?(elements, role) ||
+      violates_platform_constraint?(elements, role) ||
+      violates_exclude_platform_constraint?(elements, role) ||
+      violates_cluster_constraint?(elements, role)
+  end
+
+  # Helper to select nodes that make sense on proposal creation
+  def select_nodes_for_role(all_nodes, role, preferred_intended_role = nil)
+    # do not modify array given by caller
+    all_nodes = all_nodes.dup
```
jsuchome: Just a nitpicking - what about different name than all_nodes, when it's actually be a subset anyway...

tboerger:
LGTM, good step in the correct direction

crowbar_framework/app/models/service_object.rb
vuntz: Good point, fixed.

jsuchome:
+1

dirkmueller:
+1

PR #1209, opened 2015-01-29
Changed rubygems to HTTP

tboerger:
While development based on the gemfile i realized that a communication
throught HTTPS does not work for our bundler. It fails with
'NoMethodError: undefined method `extensions' for nil:NilClass'. With
this change it simply works again without any error.

### DIFF
--- a/crowbar_framework/Gemfile
+++ b/crowbar_framework/Gemfile
@@ -15,7 +15,7 @@
 # limitations under the License.
 #
 
-source "https://rubygems.org"
+source "http://rubygems.org"
 
 gem "rails", "~> 4.1"
 gem "haml-rails", "~> 0.5"


vuntz:
Isn't this wrong, though? Surely it's a good idea to check that the server is really the one we think it is?

tboerger:
@vuntz Feel free to make a better fix :)
This is used in development/test only anyway.

vuntz:
Can't make a better fix since I have no clue how to fix it, but I guess someone needs to look at bundler?

tboerger:
Now it's getting really strange... For the Gemfiles on the other barclamps it works with https.

bkutil:
@tboerger where & how did you get this error? Is this PR still valid?

tboerger:
If it comes up again i will reopen it.

PR #1210, opened 2015-01-29
Stop overwriting /etc/sudoers on the admin node

dirkmueller:
We really only need to add one line, so try to do
that instead


tboerger:
:+1:

dirkmueller:
finished testing, seems to be working.

chef/cookbooks/crowbar/recipes/default.rb
```diff
@@ -141,6 +141,30 @@
...
+  owner "root"
+  group "root"
+  mode "0750"
+  action :create
+end
+
+ruby_block "Ensure /etc/sudoers.d is included in sudoers" do
+  block do
+    f = Chef::Util::FileEdit.new("/etc/sudoers")
+    f.insert_line_if_no_match(/^#includedir \/etc\/sudoers.d/,
```
vuntz: I guess adding a `$` at the end of the regexp would be a tiny bit more secure.

But other than that, +1

dirkmueller: @vuntz my main thought was that there might be a "#include/sudoers.d/" already, and thats also fine. we can make the regexp /?$ though

PR #1211, opened 2015-01-29
Fix undefined variable exception

dirkmueller:
I guess this would have been found if we'd have test
coverage for this..


vuntz:
Ouch. +1

PR #1213, opened 2015-01-29
Respect count constraint in select_nodes_for_role

vuntz:


crowbar_framework/app/models/service_object.rb
```diff
@@ -669,6 +669,10 @@ def select_nodes_for_role(all_nodes, role, preferred_intended_role = nil)
       valid_nodes = preferred_all_nodes unless preferred_all_nodes.empty?
     end
 
+    if role_constraints[role] && role_constraints[role].has_key?("count")
+      valid_nodes = valid_nodes.take(role_constraints[role]["count"])
```
dirkmueller: according to http://www.ruby-doc.org/core-2.2.0/Array.html#method-i-take take does an exception on negative values. please check for >=0 first.

tboerger: Simply  if role_constraints[role] && role_constraints[role]["count"]

vuntz: @tboerger what's the difference? (and, well, why do we have this nice method that is readable? :-))

vuntz: @dirkmueller done

tboerger: Because we just need to call it if the value is above 1? ```role_constraints[role].has_key?("count")``` is even for values like ```-1``` true because it just checks for the key and not the value like my solution.

tboerger: Your solution just adds one more more or less useless check.

vuntz: We want `-1` to avoid entering the code path, but that doesn't work:

```ruby
irb(main):005:0> puts "bug" if -1
bug
=> nil
```

tboerger: That's odd... Negative integers have to be a false value -.-

bkutil:
+1

PR #1214, opened 2015-01-30
Display not 100% available OS as disabled in dropdown menu

vuntz:
Depends on https://github.com/crowbar/barclamp-provisioner/pull/375

### DIFF
--- a/crowbar_framework/app/helpers/form_helper.rb
+++ b/crowbar_framework/app/helpers/form_helper.rb
@@ -19,7 +19,8 @@ module FormHelper
   def platforms_for_select(selected)
     options_for_select(
       available_platforms.map { |p| [crowbar_service.pretty_target_platform(p), p] },
-      selected.to_s
+      selected: selected.to_s,
+      disabled: disabled_platforms
     )
   end
 
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -108,6 +108,19 @@ class NodeObject < ChefObject
     end
   end
 
+  def self.disabled_platforms
+    @disabled_platforms ||= begin
+      provisioner = NodeObject.find("roles:provisioner-server").first
+      if provisioner.nil?
+        []
+      else
+        provisioner["provisioner"]["available_oses"].keys.select { |p|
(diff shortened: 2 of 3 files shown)


crowbar_framework/app/models/node_object.rb
```diff
@@ -74,7 +74,7 @@ def self.default_platform
   end
 
   def self.available_platforms
-    @@available_platforms ||= begin
+    @available_platforms ||= begin
```
tboerger: Are you sure that you have to remove the second ```@```? I think that is only the case for ```self << class```

vuntz: I tried, and yes, it was needed.

crowbar_framework/app/helpers/form_helper.rb
```diff
@@ -19,7 +19,8 @@ module FormHelper
   def platforms_for_select(selected)
     options_for_select(
       available_platforms.map { |p| [crowbar_service.pretty_target_platform(p), p] },
-      selected.to_s
+      selected: selected.to_s,
+      disabled: disabled_platforms.map { |p| p }
```
tboerger: Why the map?

vuntz: Just some over-aggressive copy&paste :-) Fixed

dirkmueller:
@vuntz looks like tests need to be adjusted, could you check?

crowbar_framework/app/models/node_object.rb
```diff
@@ -485,6 +498,10 @@ def allocate!
     return if @node.nil?
     return if @role.nil?
     return if self.allocated?
+
+    raise "Platform #{self.target_platform} is not fully setup for installation yet." if NodeObject.disabled_platforms.include? self.target_platform
```
tboerger: Maybe swap this? Unavailable platforms are more important :)

vuntz:
@dirkmueller I'm removing the commit that fails the tests for now. Will resubmit (and fix tests) in a later PR so we can go ahead and merge this if we want it now.

vuntz:
Anyone up for a review?

tboerger:
:+1:

PR #1216, opened 2015-01-31
documentation for crowbar batch

aspiers:
https://trello.com/c/8ntaSRJ4

### DIFF
--- a/crowbar_framework/doc/default/crowbar/batch.md
+++ b/crowbar_framework/doc/default/crowbar/batch.md
@@ -1,0 +1,142 @@
+# `crowbar-batch`
+
+This is the documentation for the
+[`crowbar batch`](../bin/crowbar_batch) subcommand.
+
+## Description
+
+As the name suggests, `crowbar batch` is intended to be run in "batch
+mode", i.e. mostly unattended, and can be used to accurately capture
+the configuration of an existing Crowbar environment, or drive Crowbar
+to build a complete new environment from scratch.
+
+`crowbar batch` has two modes of operation:
+
+*   `crowbar batch export`
+
+    Exports a YAML file which describes existing proposals and how
+    their parameters deviate from the default proposal values for
+    that barclamp.
+
+*   `crowbar batch build`
+
+    Imports a YAML file in the same format as above, and uses it to
+    build new proposals if they don't yet exist, and then update the
+    existing proposals so that their parameters match those given in
+    the YAML file.
+
+## YAML file format
+
+Here is an example YAML file.  At the top-level, there is a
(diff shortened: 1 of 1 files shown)


vuntz:
+1

tboerger:
If you integrate it into ```crowbar_framework/doc``` it gets automaticly integrated into the upstream online documentation within the crowbar UI.

aspiers:
@tboerger AFAIK that only became true when they implemented it in the early days of Crowbar 2, and I don't see any evidence of that anywhere. Please can you give more details?  What tree structure are you talking about?

tboerger:
You have never seen that because you are not working in the git based source. Our RPM's are replacing the full upstream documentation that got revamped by me for stoney. The markdown files added to ```crowbar_framework/doc``` folder and to ```crowbar_framework/doc/crowbar.yml``` are getting integrated into the help page within crowbar.

tboerger:
https://github.com/crowbar/barclamp-crowbar/blob/master/crowbar_framework/app/controllers/docs_controller.rb

aspiers:
OK so it's still totally unfinished according to https://github.com/crowbar/barclamp-crowbar/tree/master/crowbar_framework/doc/default/crowbar - I looked at that and assumed it wasn't used at all.

tboerger:
Of course it have to be improved as currently nobody takes care of it. But i think it's a nice place to start :)

aspiers:
This is one of several areas where we are reinventing the wheel thanks to the fork of OpenCrowbar :-(  This was done in Crowbar 2 a long time ago.  Oh well ...

tboerger:
Feel free to migrate to opencrowbar :P

dirkmueller:
so from what I can see we can still merge this as is, right? LGTM

aspiers:
I've just rebased and moved the `batch.md` file to be under `crowbar_framework/doc/default/crowbar/` as suggested by @tboerger; however I didn't update `crowbar_framework/doc/crowbar.yml` because I don't know how that's supposed to work.  We can do that later if it makes sense.

So yes I think this should be fine to merge now.

tboerger:
Fine with me. Hopefully we will get back to better docs generator later. :+1:

aspiers:
I'm just wondering if it's OK to violate DRY here, considering the same doc is now in the deployment guide:

http://docserv.nue.suse.com/documents/Cloud5/susecloud-deployment/html/app.deploy.crowbatch.html

Is it worth having the same thing in two locations?

vuntz:
@aspiers this is really upstream doc vs downstream doc. It's worth it.

aspiers:
OK then :)

PR #1217, opened 2015-02-03
Treat DMI information as optional

dirkmueller:
Some (new) hardware is not supported by dmidecode, so the
dmi tree is missing in ohai. Just ignore it, since the
whole point of that is logging the information.

### DIFF
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -58,7 +58,7 @@ class NodeObject < ChefObject
     elsif nodes.length == 0
       nil
     else
-      raise "#{I18n.t('multiple_node_alias', :scope=>'model.node')}: #{nodes.join(',')}"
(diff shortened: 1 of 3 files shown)


tboerger:
IMHO this logging line is anyway useless but i'm fine with this change.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1335,7 +1335,8 @@ def unique_device_for(device)
 
       # VirtualBox does not provide stable disk ids, so we cannot rely on them
       # in that case.
-      unless @node[:dmi][:system][:product_name] =~ /VirtualBox/i
+      dmisystem=@node[:dmi][:system][:product_name] rescue "unknown"
```
vuntz: Style nitpick: missing spaces around `=`

vuntz:
+1

tboerger:
LGTM

PR #1218, opened 2015-02-03
Fix RadosGW on SLE12

dirkmueller:
The deployed RadosGW configuration was not compatible with Apache 2.4,
which is to be used on SLE12.

### DIFF
--- a/chef/cookbooks/apache2/templates/suse/mods/fastcgi.conf.erb
+++ b/chef/cookbooks/apache2/templates/suse/mods/fastcgi.conf.erb
@@ -3,8 +3,12 @@
 	<Directory "/srv/www/fcgi-bin">
 		AllowOverride None
 		Options None
-		Order allow,deny
+		<%- if node[:apache][:version].to_f < 2.4 %>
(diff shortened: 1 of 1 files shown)


chef/cookbooks/apache2/templates/suse/mods/fastcgi.conf.erb
```diff
@@ -3,8 +3,12 @@
 	<Directory "/srv/www/fcgi-bin">
 		AllowOverride None
 		Options None
-		Order allow,deny
+                <%- if node[:apache][:version].to_f < 2.4 %>
+                Order deny,allow
 		Deny from all
+                <%- else %>
+                Deny from all
```
vuntz: Should be `Require all denied`

vuntz: Also: tabs vs spaces (this file uses tabs everywhere)

vuntz:
+1

tboerger:
+1

PR #1219, opened 2015-02-04
Add some states to list of valid transition states

vuntz:
sledgehammer and crowbar_join need to be able to transition states to:

 - problem: errors on boot
 - reboot: on rebooting
 - shutdown: on shutting down
 - debug: special state for sledgehammer that can be triggered for
   debugging


tboerger:
:+1:

dirkmueller:
+1

PR #1220, opened 2015-02-04
Allow direct calls to crowbar_reset

vuntz:
As we put this in bin now, being able to directly call it instead of
displaying some bug message helps.


tboerger:
LGTM

bkutil:
+1

PR #1221, opened 2015-02-05
Ensure that sudo is installed

dirkmueller:


vuntz:
+1

chef/cookbooks/crowbar/recipes/default.rb
```diff
@@ -49,6 +49,7 @@
 when "suse"
   pkglist = %w(
     curl
+    sudo
```
vuntz: Actually, can you add sudo to the list of packages for the other distros too?

dirkmueller: we removed support for the other distros..

PR #1222, opened 2015-02-05
Extend target platform with suse-12.0 for method unique_device_for

mjura:


vuntz:
I thought this was on purpose. @rhafer?

rhafer:
Yes, this was on purpose. SLES12 doesn't create the by-path links for virtio devices (unless that changed again). See also the comment above the code block. So -1 on this.

mjura:
Hmmm I see, thats way `unclaimed_physical_drives` doesn't work on SLE12

PR #1224, opened 2015-02-10
delete _remove roles from the run_list

MaximilianMeister:


MaximilianMeister:
@vuntz I haven't tested it yet. in terms of databag, is that more or less what you had in mind?

MaximilianMeister:
@vuntz can you take another look please? previous discussion was in https://github.com/crowbar/barclamp-crowbar/issues/1162

crowbar_framework/app/models/service_object.rb
```diff
@@ -1173,6 +1173,10 @@ def apply_role(role, inst, in_queue)
 
       nodes_in_batch = []
 
+      # get databag to remember potential removal of a role
+      databag = Chef::DataBag.load "barclamps/#{@bc_name}"
+      databag["crowbar"]["roles_to_remove"] = []
```
vuntz: It might have been my idea, so if it is, I apologize :-) But I think just adding an attribute in the data bag like this will break validation when we save a proposal that still has this attributes.

There might be a clever/hacky way to handle this, though (by putting the attributes under `elements`, maybe?).

crowbar_framework/app/models/service_object.rb
```diff
@@ -1401,6 +1408,15 @@ def apply_role(role, inst, in_queue)
     # XXX: This should not be done this way.  Something else should request this.
     system("sudo", "-i", Rails.root.join("..", "bin", "single_chef_client.sh").expand_path.to_s) if !ran_admin
 
+    databag = Chef::DataBag.load "barclamps/#{@bc_name}"
```
vuntz: Why are we reloading the databag?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1186,6 +1190,9 @@ def apply_role(role, inst, in_queue)
           tmprole = RoleObject.find_role_by_name "#{role_name}_remove"
           unless tmprole.nil?
             elem_remove = tmprole.name
+            # save remove intention in #{@bc_name}-databag
+            databag["crowbar"]["roles_to_remove"] << "#{role_name}_remove"
```
vuntz: Hrm: shouldn't we save for which nodes the role will be removed? Otherwise, I'm not sure it's very useful. My main worry was that we lose the reference of the nodes that are removed when apply_role fails somehow. So not the reference of the roles, but really the nodes.

crowbar_framework/app/models/service_object.rb
MaximilianMeister: @vuntz good catch. i guess i can reach that `elements` over the `ProposalObject`.
but this will be dropped soon IIUC.
shall I go the way over the `ProposalObject` for now?

MaximilianMeister: we could also add validations, but this would be a fixed attribute then just for that one case (which likely rarely happens)

MaximilianMeister: > i guess i can reach that elements over the ProposalObject.

ah i just saw `elements` is in the crowbar databag... sorry

MaximilianMeister:
@vuntz the information is now in the right databag, validations should not be affected anymore

MaximilianMeister:
any updates on this?
@vuntz would be good if we could get this done this sprint as we carried it over some sprints already
@aspiers @tboerger @bkutil are invited to review as well of course if you find some time :)
thanks

bkutil:
Codewise, this looks good to me.

aspiers:
I haven't looked at the code in detail yet, but at a glance my only criticism is the way the commits are split.  Huge kudos for adding nice tests (yay!), but they should be in the same commit as the code they are testing.  Also the introduction of a new function should be in the same commit as its first consumer(s), otherwise at a certain point in the history, the codebase contains a totally unused function.

https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes

But at least breaking the PR up into too *many* commits is much much better than breaking it up into too *few* commits ;-)  Most people make the opposite mistake ...

MaximilianMeister:
> Also the introduction of a new function should be in the same commit as its first consumer(s), otherwise at a certain point in the history, the codebase contains a totally unused function.

Okay. My thought behind the ordering has been adding commits in an order to make as sure as possible that the whole code won't break after each commit.

> But at least breaking the PR up into too many commits is much much better than breaking it up into too few commits ;-) Most people make the opposite mistake ...

I'll change the splitting according to your suggestion to see for myself but I will surely end up with fewer commits in the end ;-)
Thanks for the review

MaximilianMeister:
@aspiers I had a look at it and the 2 functions that I added are so close together (dependent), that I would end up with only one / max 2 commits in the end. (if i split out the databag part)
I cannot call `delete_remove_roles_from_the_runlist` without calling `save_node_state` immediately afterwards. It has to be in the same commit. The specs would go in this commit as well.

aspiers:
> Okay. My thought behind the ordering has been adding commits in an order to make as sure as possible that the whole code won't break after each commit.

That should *always* be true.  But I am not proposing anything which would cause breakage.

> I'll change the splitting according to your suggestion to see for myself but I will surely end up with fewer commits in the end ;-)

Sure - like I already said, I think this PR currently has too many commits, which is better than too few.

> @aspiers I had a look at it and the 2 functions that I added are so close together (dependent)

Proximity does not imply logical dependency, nor does logical dependency of change B on change A require them to be in the same commit.  A PR could have a commit which adds some code and a subsequent commit which changes some of that code (not just code near it) and that could still potentially follow the golden rule of one logical change per commit.  For example a PR could contain a commit X which adds docstrings / comments to the headers of a bunch of functions or methods, and a subsequent commit Y which changes the API (and hence docstring / comment header) of one of those functions / methods.  Both X and Y are logically atomic, but Y depends on X because it changes some of the lines which X introduced.  And that's fine.

> that I would end up with only one / max 2 commits in the end. (if i split out the databag part)  I cannot call delete_remove_roles_from_the_runlist without calling save_node_state immediately afterwards. It has to be in the same commit. The specs would go in this commit as well.

Sounds fine to me.

MaximilianMeister:
> A PR could have a commit which adds some code and a subsequent commit which changes some of that code

@aspiers Okay, I just thought it is bad practice to change code that were introduced by an earlier commit in the same PR. Thanks for clarifying.
I hope it's all good now in terms of commit practice. Please have another look.

aspiers:
> Okay, I just thought it is bad practice to change code that were introduced by an earlier commit in the same PR. 

No, it's bad practice for a PR to introduce *breakage* in one commit and then fix it in a later commit.  Don't over-generalize that rule ;-)

> I hope it's all good now in terms of commit practice. Please have another look.

Looks a lot better, thanks!  And after the regrouping, suddenly it makes sense to me; even I'm surprised at suddenly how much easier it is for me to understand and review.  So I guess my commit style pedantry was justified after all ;-)

MaximilianMeister:
> What happens if there are multiple _remove roles and the run fails in the middle of one of them? Shouldn't each one be removed from the runlist and to_remove queue immediately after it is completed? This would reduce failure windows in the orchestration workflow.

all `_remove` roles will be in `pending_node_actions` for example

``` ruby
pending_node_actions = { node_name => { :remove => [], :add => ["heat-server_remove", "keystone-server_remove"] } }
```

`delete_remove_roles_from_runlist` iterates over them, but doesn't save the node state after each iteration. that means it is not completed until the method returns and hands the bulk over to `save_node_state`.
Only after that it is safe to drop `to_remove` from the databag.

I would have to rewrite `save_node_state` to not bulk save, but save one node at a time, then we could safely drop each `_remove` role from the databag **during** `delete_remove_roles_from_runlist`.
The specs would have to be changed as well.

Would that make sense to you? I like the idea

aspiers:
IIUC it sounds sensible to me.  @vuntz what do you think?

MaximilianMeister:
@aspiers I made some overall changes now.
* `save_node_state` saves just one node
* `_remove` roles will be dropped from the databag right after they are removed from the node's runlist (for reducing the failure windows during orchestration)
* `delete_remove_roles_from_runlist` reads directly from the databag now, instead of `pending_node_actions`. 

I am only not sure if I should take the `crowbar/bc-template-#{@bc_name}` or the `crowbar/bc-#{bc_name}-default` databag to store that temporary information about removal...

rsalevsky:
tested works :+1:

MaximilianMeister:
@tboerger @vuntz @aspiers @bkutil May I kindly ask you to have another look here? :-)

I am still not so sure which databag we should use here. `bc-template-#{@bc_name}` works, but I think it is misused. If we use `bc-#{@bc_name}-default`, we wont cover e.g. `pacemaker` which doesn't have this default databag

jdsn:
@tboerger @vuntz @aspiers @bkutil -  PLEASE, help team green to finally finish this PR and the corresponding trello card. Please review this PR and if possible merge it.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1418,6 +1440,48 @@ def apply_role(role, inst, in_queue)
...
+    end
+
+    # delete remove intention from databag, as it has been done successfully
+    if databag["deployment"][@bc_name]["elements"]["to_remove"].empty?
+      databag["deployment"][@bc_name]["elements"].delete("to_remove")
+      databag.save
+    end
+  end
+
+  def save_node_state(node, origin)
```
aspiers: I would prefer `#save_node_state` to automatically calculate the caller's origin itself by introspecting the stack trace.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1186,6 +1189,19 @@ def apply_role(role, inst, in_queue)
           tmprole = RoleObject.find_role_by_name "#{role_name}_remove"
           unless tmprole.nil?
             elem_remove = tmprole.name
+            # save remove intention in #{@bc_name}-databag
+            if databag["deployment"][@bc_name]["elements"]["to_remove"].nil?
+              databag["deployment"][@bc_name]["elements"]["to_remove"] = {}
```
aspiers: Maybe `||= {}` is cleaner?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1186,6 +1189,19 @@ def apply_role(role, inst, in_queue)
...
+            end
+
+            old_nodes.each do |old_node|
+              @logger.debug "saving #{elem_remove} intention for #{old_node}"
+              if databag["deployment"][@bc_name]["elements"]["to_remove"][old_node].nil?
```
aspiers: Maybe use a temporary variable to avoid repetition?

    to_remove = databag["deployment"][@bc_name]["elements"]["to_remove"]
    to_remove[old_node] ||= []
    to_remove[old_node] << elem_remove

aspiers: Or maybe even

    (to_remove[old_node] ||= []) << elem_remove

crowbar_framework/app/models/service_object.rb
```diff
@@ -1418,6 +1440,48 @@ def apply_role(role, inst, in_queue)
...
+      # delete nodes from databag with no more _remove roles
+      if databag["deployment"][@bc_name]["elements"]["to_remove"][node_name].empty?
+        databag["deployment"][@bc_name]["elements"]["to_remove"].delete(node_name)
+        databag.save
+      end
+    end
+
+    # delete remove intention from databag, as it has been done successfully
+    if databag["deployment"][@bc_name]["elements"]["to_remove"].empty?
+      databag["deployment"][@bc_name]["elements"].delete("to_remove")
```
aspiers: Why is this necessary? Won't it just get recreated again later?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1418,6 +1440,48 @@ def apply_role(role, inst, in_queue)
     [200, {}]
   end
 
+  def delete_remove_role_from_runlist(nodes_with_roles_to_remove)
```
aspiers: This method is incorrectly named - the name implies that it only removes one role from one runlist, but actually it will remove multiple roles from multiple runlists.  Better to call it `delete_remove_roles_from_runlists` and have the body of the inner loop call `delete_remove_role_from_runlist`.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1418,6 +1440,48 @@ def apply_role(role, inst, in_queue)
...
+        save_node_state(node, __method__)
+
+        # delete _remove roles from databag
+        databag["deployment"][@bc_name]["elements"]["to_remove"][node_name].delete(role_to_remove)
+        databag.save
+      end
+
+      # delete nodes from databag with no more _remove roles
+      if databag["deployment"][@bc_name]["elements"]["to_remove"][node_name].empty?
+        databag["deployment"][@bc_name]["elements"]["to_remove"].delete(node_name)
```
aspiers: Why is this necessary? Won't it just get recreated again later?

aspiers:
Looks mostly good to me - still a few minor points.

crowbar_framework/app/models/service_object.rb
MaximilianMeister: as we are (ab)using the `bc-template-#{@bc_name}` databag I would like to get it back to how it was before. this is the hacky solution that I came up with by @vuntz suggestion [here](https://github.com/crowbar/barclamp-crowbar/pull/1224#discussion_r27341071)

otherwise if there is just **one** point in time, where some role has to be removed, we will have this `to_remove` attribute in the databag under `elements` forever, even though it may not be used anymore.

So this is needed as a cleanup.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1190,6 +1193,15 @@ def apply_role(role, inst, in_queue)
           tmprole = RoleObject.find_role_by_name "#{role_name}_remove"
           unless tmprole.nil?
             elem_remove = tmprole.name
+            # save remove intention in #{@bc_name}-databag
+            databag["deployment"][@bc_name]["elements"]["to_remove"] ||= {}
+
+            old_nodes.each do |old_node|
+              @logger.debug "saving #{elem_remove} intention for #{old_node}"
+              databag["deployment"][@bc_name]["elements"]["to_remove"][old_node] ||= []
+              databag["deployment"][@bc_name]["elements"]["to_remove"][old_node] << elem_remove
```
MaximilianMeister: @aspiers I had to revert some of the changes that reduce duplication like 
* `to_remove = databag["deployment"][@bc_name]["elements"]["to_remove"]`
* `(to_remove[old_node] ||= []) << elem_remove`

the reference to the databag object gets lost (to_remove doesn't have it anymore), and subsequently things don't work anymore. That's why I reverted it

crowbar_framework/app/models/service_object.rb
```diff
@@ -1168,6 +1168,9 @@ def apply_role(role, inst, in_queue)
     # of names of Chef roles.
     run_order = []
 
+    # get databag to remember potential removal of a role
+    databag = Chef::DataBag.load "crowbar/bc-template-#{@bc_name}"
```
vuntz: Do you really mean the template? Or did you want the proposal?

MaximilianMeister: I asked this a few times already but I got no answer...
Thanks for the hint, I will use the proposal then :)

bkutil:
+1 Looks good to me. @vuntz what do you think?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1168,6 +1168,9 @@ def apply_role(role, inst, in_queue)
     # of names of Chef roles.
     run_order = []
 
+    # get databag to remember potential removal of a role
+    databag = ProposalObject.find_proposal(@bc_name, inst)
```
vuntz: Shouldn't we read from the databag to also populate `old_nodes`? I mean, the goal for this hack is to make sure chef-client is run for the removed nodes in case there's a crash. And based on the code here, I don't see that happening -- or am I missing something?

MaximilianMeister: no you are right, we need sth like
`old_nodes = old_elements[role_name] || nodes_from_to_remove`

crowbar_framework/app/models/service_object.rb
```diff
@@ -1178,7 +1181,7 @@ def apply_role(role, inst, in_queue)
       nodes_in_batch = []
 
       elems.each do |role_name|
-        old_nodes = old_elements[role_name]
+        old_nodes = old_elements[role_name] || databag["deployment"][@bc_name]["elements"]["to_remove"].keys
```
MaximilianMeister: @vuntz this would give us any old_nodes that were left over after a crash, but just if apply role runs in the context of this barclamp... do you think that is enough? or should we gather the leftover nodes from **all** barclamps?

vuntz: I think "only this barclamp" is fine.

MaximilianMeister:
@vuntz I just saw we have a validation error if we keep it like this

```
"elements": {
    "heat-server": [
      "d52-54-01-77-77-01.emil.suse.de"
    ],
    "to_remove": { # <<< this can only be an array / sequence
      "d52-54-01-77-77-01.emil.suse.de": [
        "heat-server_remove"
      ]
    }
  },
```

I'll have to update it once more...

vuntz:
@MaximilianMeister what about doing this instead?

```
"elements": {
    "heat-server": [
      "d52-54-01-77-77-01.emil.suse.de"
    ],
    "heat-server_remove": [
      "d52-54-01-77-77-01.emil.suse.de"
    ]
  },
```

MaximilianMeister:
@vuntz done

crowbar_framework/app/models/service_object.rb
aspiers: @MaximilianMeister I don't understand - the `databag` variable still refers to the databag object, so why do you say it "gets lost"?

MaximilianMeister: `databag.save` in the end doesn't save the changes that happen during:
`(to_remove[old_node] ||= []) << elem_remove`
so I meant `to_remove` doesn't reference to `databag`.

aspiers: `to_remove` isn't supposed to reference the whole databag, just a small part of it.  Why would that matter?  Whether you write `to_remove` or `databag["deployment"][@bc_name]["elements"]["to_remove"]`, shouldn't you get the same object every time?

MaximilianMeister: apparently not. because when it comes to `databag.save` the changes are not in `databag`. so it is a different object.

aspiers: Of course it's a different object!  After all we didn't write

    to_remove = databag

But that still doesn't explain how there could be different behaviour.  I challenge you to find a value of `hash` for which

    hash['foo']['bar'] = 'baz'

and

    nested_hash = hash['foo']
    nested_hash['bar'] = 'baz'

behave differently :-)  In both cases a lookup is done on the top-level object, returning another object, and then assignment is done on the second object.

bkutil:
+1

crowbar_framework/app/models/service_object.rb
```diff
@@ -1190,6 +1195,16 @@ def apply_role(role, inst, in_queue)
...
           unless tmprole.nil?
             elem_remove = tmprole.name
+            # save remove intention in #{@bc_name}-databag
+            databag["deployment"][@bc_name]["elements"][elem_remove] ||= []
+
+            old_nodes.each do |old_node|
+              @logger.debug "saving #{elem_remove} intention for #{old_node}"
+              unless databag["deployment"][@bc_name]["elements"][elem_remove].include? old_node
+                databag["deployment"][@bc_name]["elements"][elem_remove] << old_node
+                databag.save
```
vuntz: Can we have some `save_databag` variable that we use to only save the databag once, at the end of all the loops?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1405,6 +1420,17 @@ def apply_role(role, inst, in_queue)
     # XXX: This should not be done this way.  Something else should request this.
     system("sudo", "-i", Rails.root.join("..", "bin", "single_chef_client.sh").expand_path.to_s) if !ran_admin
 
+    # are there any roles to remove from the runlist?
+    role_to_remove = databag["deployment"][@bc_name]["elements"].select {|r| r =~ /_remove/}.keys.first
```
vuntz: Doesn't that assume that there's only one remove role? What about barclamps that have multiple roles, which, I would expect to also have multiple remove roles?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1180,6 +1184,7 @@ def apply_role(role, inst, in_queue)
       elems.each do |role_name|
         old_nodes = old_elements[role_name]
         new_nodes = new_elements[role_name]
+        leftover_nodes = databag["deployment"][@bc_name]["elements"]["#{role_name}_remove"]
```
vuntz: Shouldn't we add stuff to `leftover_nodes`, since there can be multiple roles and we don't want to override the value?

vuntz: Or actually, you probably want a hash and do `leftover_nodes["#{role_name}_remove"] = ...`

crowbar_framework/app/models/service_object.rb
MaximilianMeister: I tested this with `heat` which has only one role. I will test it with `ceilometer` and update the PR accordingly

crowbar_framework/app/models/service_object.rb
MaximilianMeister: Good point but I'd rather keep it this way, in case a crash happens here in the loop we would have an inconsistent state in the databag

vuntz: I'm not sure I see what would be inconsistent. Are you thinking of something specific?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1428,6 +1454,35 @@ def apply_role(role, inst, in_queue)
     chef_daemon(:start, chef_daemon_nodes)
   end
 
+  def delete_remove_role_from_runlist(node_name, role_to_remove, inst)
```
MaximilianMeister: this constraint will hopefully be relaxed in the rubocop conf.

crowbar_framework/app/models/service_object.rb
MaximilianMeister: @vuntz I can save it once outside the loop, but i would do that without an extra variable. i cannot imagine a condition where we wouldn't want to save it.

MaximilianMeister:
@vuntz I updated the missing bits, please have another look
I also added some comments to be a bit more verbose what happens.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1193,6 +1205,17 @@ def apply_role(role, inst, in_queue)
...
+            databag["deployment"][@bc_name]["elements"][elem_remove] ||= []
+
+            old_nodes.each do |old_node|
+              @logger.debug "saving #{elem_remove} intention for #{old_node}"
+              unless databag["deployment"][@bc_name]["elements"][elem_remove].include? old_node
+                databag["deployment"][@bc_name]["elements"][elem_remove] << old_node
+              end
+            end
+
+            databag.save
```
vuntz: @bkutil will saving like this trigger a bump in the revision, which would result in making the proposal not appear as applied for the latest revision?

vuntz: @MaximilianMeister re my earlier comment about less saving: I guess my suggestion was to save only once we reach the end of the `element_order.each do |elems|` loop. Right now, you save once per role (in worst case), instead of just once.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1408,6 +1431,35 @@ def apply_role(role, inst, in_queue)
...
+    roles_to_remove = databag["deployment"][@bc_name]["elements"].select do |r|
+      r =~ /_remove/
+    end.keys
+    # returns ["role1_remove", "role2_remove"] || {}
+    roles_to_remove.each do |role_to_remove|
+      nodes_with_role_to_remove = databag["deployment"][@bc_name]["elements"][role_to_remove]
+      # returns ["node1", "node2"] || nil
+      # are there any leftover nodes where runlist removal
+      # has not been performed?
+      unless nodes_with_role_to_remove.nil?
```
vuntz: I don't think this can be `nil`, as otherwise, the `select` above wouldn't have returned role. Or can it still be `nil`?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1408,6 +1431,35 @@ def apply_role(role, inst, in_queue)
...
+        # concatenate the leftover_nodes with
+        # the actual nodes that have a _remove role
+        unless leftover_nodes[role_to_remove].nil?
+          nodes_with_role_to_remove.concat(leftover_nodes[role_to_remove])
+        end
```
vuntz: I think this is already done at https://github.com/crowbar/barclamp-crowbar/pull/1224/files#diff-0be8ea1f05c9441a59bb2b7ac65b1c62R1196

I'd actually rewrite the whole block you're adding to:

```ruby
leftover_nodes.each do |role_to_remove, nodes_with_role_to_remove|
  nodes_with_role_to_remove.each do |node_name|
    nodes_with_role_to_remove.each do |node_name|
  end
end
```

MaximilianMeister: @vuntz I am a bit confused. I thought the `leftover_nodes` are from a former `apply_role` which hasn't done the removal yet for some reason. So if we have an `apply_role` run without `leftover_nodes` but with a "new" removal, it won't happen, because the nodes added [here](https://github.com/crowbar/barclamp-crowbar/pull/1224/files#diff-0be8ea1f05c9441a59bb2b7ac65b1c62R1214) are not in the `leftover_nodes` hash.

crowbar_framework/app/models/service_object.rb
MaximilianMeister: you are right, thanks for the catch :-)

crowbar_framework/app/models/service_object.rb
MaximilianMeister: > will saving like this trigger a bump in the revision, which would result in making the proposal not appear as applied for the latest revision?

@vuntz
do you mean the `crowbar-revision` ?
as i observed it counts up quite a few times on each apply or deactivate (12 times in my case with deactivating ceilometer). I don't know the exact logic behind it yet, all i know is that in deactivated state, it is yellow and says "has been changed since last deployment". applying it again it is green and up to date.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1183,6 +1187,14 @@ def apply_role(role, inst, in_queue)
       elems.each do |role_name|
```
vuntz: Can you add a `next if role_name =~ /_remove$/` here, just to be safe? Otherwise, if someone manually adds a _remove role in the element_order attribute, then we will have some fun stuff happening.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1408,6 +1432,33 @@ def apply_role(role, inst, in_queue)
...
 
+    # are there any roles to remove from the runlist?
+    # The @bcname proposal's elements key will contain the removal intentions
+    # proposal.elements =>
+    # {
+    #   "role1_remove" => ["node1"],
+    #   "role2_remove" => ["node2", "node3"]
+    # }
+    roles_to_remove = databag["deployment"][@bc_name]["elements"].select do |r|
+      r =~ /_remove/
```
vuntz: Maybe `/_remove$/`?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1235,6 +1256,9 @@ def apply_role(role, inst, in_queue)
       @logger.debug "run_order #{run_order.inspect}"
     end
 
+    # save databag with the role removal intention
+    databag.save
```
vuntz: But we should only save if there was a change :-) That's why I suggested earlier to use a variable for this.

vuntz:
After discussion with @MaximilianMeister, moved to https://github.com/crowbar/barclamp-crowbar/pull/1307

crowbar_framework/app/models/service_object.rb
tboerger: :+1:

PR #1225, opened 2015-02-10
Refactor platform version requirements check

bkutil:
As the service object is already too bloated, this set of patches moves the node platform version checks into a separate classes.

It also simplifies the code a bit by not using lambdas, but having the methods as a part of (private) interface.

cc @aplanas

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -889,7 +889,7 @@ class ServiceObject
         node = NodeObject.find_node_by_name(element)
 
         return true if !constraints.any? do |platform, version|
-          node[:platform] == platform && equal_or_match_version(node[:platform_version], version)
+          node[:platform] == platform && PlatformRequirement.new(version).satisfied_by?(node[:platform_version])
         end
       end
     end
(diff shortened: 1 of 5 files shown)


aplanas:
... but ... but I love lambdas ...

+1

crowbar_framework/app/models/service_object.rb
```diff
@@ -889,7 +889,7 @@ def violates_platform_constraint?(elements, role)
         node = NodeObject.find_node_by_name(element)
 
         return true if !constraints.any? do |platform, version|
-          node[:platform] == platform && equal_or_match_version(node[:platform_version], version)
+          node[:platform] == platform && PlatformRequirement.new(version).satisfied_by?(node[:platform_version])
```
aspiers: Could we also add a helper on the `node` object to tidy this up?

    node[:platform] == platform && node.version_matches(version)

aspiers:
Other than my suggestion above, +100 for this kind of refactoring ;-)  I hate bloat!

bkutil:
@aplanas argh, sorry :( I promise to add some lambdas back in the future, to restore the balance in the functional universe.

tboerger:
:+1:
Forget the lambdas :)

PR #1226, opened 2015-02-11
Display better error for invalid groups

bkutil:
When bulk editing nodes, currently the default error message is
displayed, when the group contains invalid characters.

 'Failed to save the nodes X.'

This patch adds correct handling of such case and the UI will
display a more informative message:

 'The group should start ...'

cc @tboerger @vuntz

### DIFF
--- a/crowbar_framework/app/controllers/nodes_controller.rb
+++ b/crowbar_framework/app/controllers/nodes_controller.rb
@@ -75,7 +75,8 @@ class NodesController < ApplicationController
       :success => [],
       :failed => [],
       :duplicate_public => false,
-      :duplicate_alias => false
+      :duplicate_alias => false,
+      :group_error => false,
     }.tap do |report|
       node_values = params[:node] || {}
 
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/controllers/nodes_controller.rb
```diff
@@ -178,6 +180,8 @@ def bulk
         "nodes.list.duplicate_alias"
       when @report[:duplicate_publics]
         "nodes.list.duplicate_public"
+      when @report[:group_error]
+        "nodes.list.group_error"
```
vuntz: This won't work as this string requires passing `:node` in `I18n.t`.

tboerger:
Overall this code really sucks. But i'm fine with this fix of error message for now.

crowbar_framework/app/controllers/nodes_controller.rb
bkutil: Fixed

vuntz:
Not tested, but +1 assuming someone tested it :-)

bkutil:
> 10:03 <cmueller> patch tested. works great

PR #1227, opened 2015-02-11
Fix uniq validation for node alias (bnc#915406)

bkutil:
A simple typo fix + test case.

!node.handle was always false, so the comparison with the new handle was
also false, and the dupe check error was never raised.

cc @tboerger @vuntz

https://bugzilla.suse.com/show_bug.cgi?id=915406


vuntz:
+1

aspiers:
LGTM

PR #1228, opened 2015-02-11
Do not allocate node if target platform is disabled / not available

vuntz:


tboerger:
:+1:

bkutil:
@vuntz maybe a helper method for the NodeObject.disabled_platform.include?(self.platform), something like `platform_available?` or `platform_enabled?` might make th e checks more readable?

bkutil:
The test fails because the platform checks fail for that particular node. You could either make sure the setup is right (available_platforms includes the node's platform), or move the checks to helpers, and then stub them to make them pass.

rsalevsky:
I would prefer to close the PR if you have not enough time to work on it.

rsalevsky:
@vuntz, this repository got merged into [crowbar-core](https://github.com/crowbar/crowbar-core).

rsalevsky:
Cloud 5 (Tex) is EOL since 15th may if this patch is still required please open a new pull request.

PR #1230, opened 2015-02-16
Added badges and rspec coverage

tboerger:


bkutil:
@tboerger looks good to me. So travis will check only pulls against master/stoney/tex ? It might be good to watch older releases, at least those which we still have to maintain.

tboerger:
> @tboerger looks good to me. So travis will check only pulls against master/stoney/tex ? It might be good to watch older releases, at least those which we still have to maintain.

That's how i understand this feature, and as far as i know roxy will be out of maintainance after release of tex.

bkutil:
Ok, thanks for info.

PR #1231, opened 2015-02-16
Show asset tag if set on the node

dirkmueller:
The asset tag is a manually configurable label for the node and
can be helpful in identifying the particular machine that
we're looking for.

### DIFF
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -563,7 +563,12 @@ class NodeObject < ChefObject
     if virtual?
       "vm-#{mac.gsub(':',"-")}"
     else
-      @node[:dmi]["chassis"]["serial_number"] rescue nil
+      asset = @node[:dmi]["chassis"]["serial_number"] rescue nil
+      if @node[:dmi]["chassis"].has_key?("asset_tag") && !@node[:dmi]["chassis"]["asset_tag"].empty?
+        "#{asset} (#{@node[:dmi]["chassis"]["asset_tag"]})"
+      else
+        asset
+      end
     end
   end
 


vuntz:
+1

crowbar_framework/app/models/node_object.rb
```diff
@@ -563,7 +563,12 @@ def asset_tag
     if virtual?
       "vm-#{mac.gsub(':',"-")}"
     else
-      @node[:dmi]["chassis"]["serial_number"] rescue nil
+      asset = @node[:dmi]["chassis"]["serial_number"] rescue nil
+      if @node[:dmi]["chassis"].has_key?("asset_tag") && !@node[:dmi]["chassis"]["asset_tag"].empty?
+        "#{asset} (#{@node[:dmi]["chassis"]["asset_tag"]})"
+      else
+        asset
+      end
```
vuntz: Actually, what about the following:

```ruby
serial = @node[:dmi]["chassis"]["serial_number"] rescue nil
asset = @node[:dmi]["chassis"]["asset_tag"] rescue nil
if asset.nil? || asset.empty?
  serial
elsif serial.nil? || serial.empty?
  asset
else
  "#{asset} (#{serial})
end
```

tboerger: I would prefer ```asset.to_s.empty?``` and ```serial.to_s.empty?```, just a single check :)

tboerger: Anf why do we mix symbol and string keys? Please use only one of that.

bkutil: `blank?` wouldn't work here?

bkutil: @tboerger what bothers me more is `a["foo"]["bar"]["baz"] rescue nil` sprinkled around *Objects. 

Wouldn't hashie's DeepFind help us here? We could have a static, declared map of "method_name" => [lookup, default value]", then loop over them and use define method? And/or use Mash + bang! methods for the setters?

Or at least we might want to split the accessors to individual 'theme' related modules (i.e RaidAttributes, NetworkAttributes).

tboerger: @bkutil my comments are related to @vuntz comment. And we can not split this informations, these are values provided by ohai.

vuntz:
Superseded by https://github.com/crowbar/barclamp-crowbar/pull/1258

PR #1232, opened 2015-02-16
Use haproxy public name for public URL when using a cluster

vuntz:
We now have CrowbarPacemakerHelper.cluster_haproxy_vpublic_name, which
enables the operator to avoid using the generated hostnames.

(cherry picked from commit e29245b490e7188196f0fca18675c8030ea24277)


tboerger:
:+1:

dirkmueller:
+1

PR #1233, opened 2015-02-16
Totally ignore nodes that were never discovered by crowbar (bnc#900216)

vuntz:
If chef-client is run on a node, but that node never went through the
discovered state, then crowbar never created the role for that node
(crowbar-$node).

This will result in failures when we list nodes because chef knows about
the node, and crowbar thinks it should know about it too.

Instead, if we see that the node role doesn't exist, then we pretend the
node just doesn't even exist.

https://bugzilla.suse.com/show_bug.cgi?id=900216

### DIFF
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -39,13 +39,8 @@ class NodeObject < ChefObject
         end
       end
       answer.compact!
-      answer.delete_if { |x| !x.has_chef_server_roles? }
     end
     return answer
-  end
-
-  def has_chef_server_roles?
-      return !@role.nil?
   end
 
   def self.find_all_nodes
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -32,8 +32,13 @@ class NodeObject < ChefObject
     if nodes.is_a?(Array) and nodes[2] != 0 and !nodes[0].nil?
       nodes[0].delete_if { |x| x.nil? }
       answer = nodes[0].map do |x|
-        NodeObject.new x
+        begin
+          NodeObject.new x
+        rescue Crowbar::Error::NotFound
+          nil
+        end
       end
+      answer.compact!
       answer.delete_if { |x| !x.has_chef_server_roles? }
     end
     return answer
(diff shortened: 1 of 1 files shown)


vuntz:
This fixes an issue that @k0da found while testing some unusual architecture: that's because crowbar_register didn't have the right key.

crowbar_framework/app/models/node_object.rb
```diff
@@ -32,8 +32,13 @@ def self.find(search)
...
       answer = nodes[0].map do |x|
-        NodeObject.new x
+        begin
+          NodeObject.new x
+        rescue Crowbar::Error::NotFound
+          nil
+        end
       end
+      answer.compact!
       answer.delete_if { |x| !x.has_chef_server_roles? }
```
vuntz: Hrm, actually, that line is supposed to deal with this already. But we don't use this `has_chef_server_roles?` method elsewhere while I guess we should.

That being said, I feel the exception is a better way to deal with this; I'd remove the `has_chef_server_roles?` bit instead.

crowbar_framework/app/models/node_object.rb
```diff
@@ -203,6 +210,7 @@ def initialize(node)
         @role = NodeObject.create_new_role(node.name, node)
       else
         Rails.logger.fatal("Node exists without role!! #{node.name}")
+        raise raise Crowbar::Error::NotFound.new
```
rhafer: ```raise raise```? Isn't that wrong or is it just my lack of ruby karma?

vuntz: Wow, fixed. I wonder how it could even work in my tests... Ah, I guess I just pasted the changes I did on a real deployment to commit to git, that would explain it.

tboerger:
LGTM

crowbar_framework/app/models/node_object.rb
```diff
@@ -134,15 +134,21 @@ def self.find_node_by_public_name(name)
...
+        NodeObject.new(chef_node)
+      else
+        nil
+      end
     rescue Errno::ECONNREFUSED => e
       raise Crowbar::Error::ChefOffline.new
     rescue StandardError => e
       Rails.logger.warn("Could not recover Chef Crowbar Node on load #{name}: #{e.inspect}")
       nil
+    rescue Crowbar::Error::NotFound => e
```
bkutil: Looks like `NotFound` is already a subclass of `StandardError`, so this rescue duplicates the one above.

bkutil:
The internal consistency of the initialized object is good, as is raising an exception if there's a problem. 

+1

Now let's see how many callers won't expect this and will fail :)

crowbar_framework/app/models/node_object.rb
vuntz: @bkutil: hrm, I don't think we want the warning in that case; I guess just moving the rescue a few lines above should help with this?

bkutil: Yep :)

vuntz: Done.

ovisan:
+1

PR #1234, opened 2015-02-16
Fix NodeObject.switch_find_info

vuntz:
The code was totally broken, setting the result variable to a value that
is not something we want to return yet, and with some totally broken
code raising an exception (info = info.name = foo, where info.name just
doesn't exist)


tboerger:
LGTM

bkutil:
@vuntz It took me a while to realize what the commit message means, i.e. that 'possibly return data from different interfaces' contains an assumption 'across calls, with different `type` arguments'. Perhaps an example / better wording might help? 

Otherwise +1

bkutil:
Oh, I see you already merged :)

PR #1235, opened 2015-02-17
Fix check for unclaimed disks (bnc#916442)

toabctl:
While checking if disks are already claimed, check first if the
current disk was already claimed and if so, use that key as unique_name
for further operations.

### DIFF
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -1321,12 +1321,47 @@ class NodeObject < ChefObject
     boot_device(device) and save
   end
 
+  # TODO: Remove duplicate code and use a gem/git submodule/whatever...
+  # see barclamp-deployer/chef/cookbooks/barclamp/libraries/barclamp_library.rb
+  def unique_name_already_claimed_by(device)
+    claimed_name = @node[:crowbar_wall][:claimed_disks].find do |claimed_name, v|
+      self.link_to_device?(device, claimed_name)
+    end || []
+    claimed_name.first
+  end
+
+  # TODO: Remove duplicate code and use a gem/git submodule/whatever...
+  # see barclamp-deployer/chef/cookbooks/barclamp/libraries/barclamp_library.rb
+  #
+  # is the given linkname a link to the given device? In attributes,
+  # that means that the given linkname can be found in the array
+  # @node[:block_device][device][by-{id,path,uuid}]
+  def link_to_device?(device, linkname)
+    # device is i.e. "sda", "vda", ...
+    # linkname is i.e. "/dev/disk/by-path/pci-0000:00:04.0-virtio-pci-virtio1"
+    lookup_and_name = linkname.gsub(/^\/dev\/disk\//, '').split(File::SEPARATOR, 2)
+    linked_devs = @node[:block_device][device][:disks][lookup_and_name[0]] rescue []
+    linked_devs.include?(lookup_and_name[1])
+  end
+
+  # IMPORTANT: keep these paths in sync with
+  # BarclampLibrary::Barclamp::Inventory::Disk#unique_name
+  # within the deployer barclamp to return always similar values.
+  # TODO: Remove duplicate code and use a gem/git submodule/whatever...
+  # see barclamp-deployer/chef/cookbooks/barclamp/libraries/barclamp_library.rb
   def unique_device_for(device)
+    # check first if we have already a claimed disk which points to the same
+    # device node. if so, use that as "unique device"
+    already_claimed_name = self.unique_name_already_claimed_by(device)
+    unless already_claimed_name.nil?
+      Rails.logger.debug("Use #{already_claimed_name} as unique_name " \
+                         "because already claimed by #{device}")
+      return already_claimed_name
+    end
+
     meta = @node["block_device"][device]
 
     if meta and meta["disks"]
-      # Keep these paths in sync with BarclampLibrary::Barclamp::Inventory::Disk#unique_name
-      # within the deployer barclamp to return always similar values.
       disk_lookups = ["by-path"]
 
       # If this looks like a virtio disk and the target platform is one


crowbar_framework/app/models/node_object.rb
```diff
@@ -1321,12 +1321,57 @@ def boot_device!(device)
     boot_device(device) and save
   end
 
+  # TODO: Remove duplicate code and use a gem/git submodule/whatever...
+  # see barclamp-deployer/chef/cookbooks/barclamp/libraries/barclamp_library.rb
+  def unique_name_already_claimed_by(device)
+    claimed_disks = @node[:crowbar_wall][:claimed_disks].to_hash
+    claimed_disks.each do | claimed_name, v |
```
bkutil: If `[:claimed_disks]` value is an array of pairs, you can skip the conversion to hash, then use `claimed_disk.find { |claimed_name, v| ... }` instead of each + return. It will also neatly return nil, if such element does not exist, making this method a one- or -two liner.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1321,12 +1321,57 @@ def boot_device!(device)
...
+  #
+  # is the given linkname a link to the given device? In attributes,
+  # that means that the given linkname can be found in the array
+  # @node[:block_device][device][by-{id,path,uuid}]
+  def link_to_device?(device, linkname)
+    # device is i.e. "sda", "vda", ...
+    # linkname is i.e. "/dev/disk/by-path/pci-0000:00:04.0-virtio-pci-virtio1"
+    meta = @node[:block_device][device]
+    if meta and meta[:disks]
+      lookup_and_name = linkname.gsub(/^\/dev\/disk\//, '').split(File::SEPARATOR)
```
bkutil: Does `linkname` value depend on `meta[:disks]` somehow (possibly being nil because it is constructed from it)? If not, it might be clearer to move this before the horrible if nest, to make it at least tiny bit more readable.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1321,12 +1321,57 @@ def boot_device!(device)
...
+      if meta[:disks].has_key(lookup_and_name[0])
+        if meta[:disks][lookup_and_name[0]].include?(lookup_and_name[1])
+          return true
```
bkutil: a simpler, branchless shortcut might be something like:

```
lookup_and_name = linkname.gsub(/^\/dev\/disk\//, '').split(File::SEPARATOR)
linked_devs = @node[:block_device][device][:disks][lookup_and_name[0]] rescue []
linked_devs.include?(lookup_and_name[1])
```
Also, can it happen that the split returns more than 2 element array? Would we care about the remaining elements? Maybe `split(File::SEPARATOR, 2)` ?

bkutil:
Looking at bugzilla and this patch, it looks good from the functionality point of view.

The problems with the code are basically stated in the TODO comments:

  * It'd be great to not add more 'generic' code to NodeObject, as it is already too big already (I realize this is not a problem introduced by this patch, but it is not helping it, either).
  * if this code is duplicating some other code in our code base, it is a de-sync waiting to happen.

If we need to merge this now, let's do that, but could you please create a Trello card mentioning that this should be refactored?

cc @vuntz

crowbar_framework/app/models/node_object.rb
toabctl: > If [:claimed_disks] value is an array of pairs, you can skip the conversion to hash, then use claimed_disk.find { |claimed_name, v| ... } instead of each + return. It will also neatly return nil, if such element does not exist, making this method a one- or -two liner.

But I don't want to have the pair as return value. I just want ```claimed_name```.

crowbar_framework/app/models/node_object.rb
toabctl: > Does linkname value depend on meta[:disks] somehow (possibly being nil because it is constructed from it)? If not, it might be clearer to move this before the if nest, to make it at least tiny bit more readable.

I don't need todo the ```gsub``` if ```meta[:disks]``` is not available. So imho I shouldn't move this code up.

crowbar_framework/app/models/node_object.rb
bkutil: Yeah, sorry. I realized that later & re-edited the comment - should've posted a new one. You can do find {} || [], then select the first element from that result.

crowbar_framework/app/models/node_object.rb
bkutil: Moving it out will make the performance only slightly bit worse, while giving a chance to refactor/make the condition more readable...

crowbar_framework/app/models/node_object.rb
toabctl: @bkutil Done in https://github.com/toabctl/barclamp-crowbar/commit/b594de6e5845828a0b5bf4d90765f3ccfbfb349f

crowbar_framework/app/models/node_object.rb
toabctl: If you think it's more readable :-) Done in https://github.com/toabctl/barclamp-crowbar/commit/b594de6e5845828a0b5bf4d90765f3ccfbfb349f

crowbar_framework/app/models/node_object.rb
toabctl: Done in https://github.com/toabctl/barclamp-crowbar/commit/b594de6e5845828a0b5bf4d90765f3ccfbfb349f .
And I haven't found a udev rule which creates deeper symlinks. So I added the ```split(File::SEPARATOR, 2)```.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1321,12 +1321,47 @@ def boot_device!(device)
...
+  # BarclampLibrary::Barclamp::Inventory::Disk#unique_name
+  # within the deployer barclamp to return always similar values.
+  # TODO: Remove duplicate code and use a gem/git submodule/whatever...
+  # see barclamp-deployer/chef/cookbooks/barclamp/libraries/barclamp_library.rb
   def unique_device_for(device)
+    # check first if we have already a claimed disk which points to the same
+    # device node. if so, use that as "unique device"
+    already_claimed_name = self.unique_name_already_claimed_by(device)
+    unless already_claimed_name.nil?
+      Rails.logger.info("Use #{already_claimed_name} as unique_name " \
```
vuntz: Move the log to debug.

toabctl: Done in https://github.com/toabctl/barclamp-crowbar/commit/d13847e408effc8d1722186fab77bdce15b9f081

PR #1236, opened 2015-02-18
Fix RemoteNode when hostname resolves to multiple addresses (bnc#918270)

vuntz:
https://bugzilla.suse.com/show_bug.cgi?id=918270


toabctl:
+1

bkutil:
+1

PR #1238, opened 2015-02-18
Fix NodeObject.unique_name_already_claimed_by if no disk claimed

vuntz:
If there's no disk claimed (like on discovery of a node), then
node[:crowbar_wall][:claimed_disks] is nil (assuming node[:crowbar_wall]
exists, which is usually not the case).


aspiers:
+1

toabctl:
+1

tboerger:
+1

PR #1239, opened 2015-02-19
Support for clusters in the queue

bkutil:
The code is unfortunately quite hairy for following reasons:

1. Cluster is just a string tag, not a separate class, or a model,
associated with nodes. So it is not possible to easily retrieve the
actual node objects, or their status, to get the status of the whole
cluster. This is visible in deploy_queue/_missing_nodes.html.haml, where
the expansion and mapping has to be done manually.

2. There aren't actual associations between objects, so it is neccessary
to query for all the objects in a Chef db, then iterate and figure out
the associations from other records.

cc @vuntz


vuntz:
+1

tboerger:
LGTM

PR #1240, opened 2015-02-19
Fix alias in bulk edit (bnc#918523)

tboerger:
If you try to set an alias or public_name on an unallocated node in bulk
edit, and you try to allocate it at the same time crowbar ignores the
alias and public_name you defined. Because of that i have moved the
alias and public_name assignment below the allocation, that way this
information doesn't get lost.

https://bugzilla.suse.com/show_bug.cgi?id=918523

### DIFF
--- a/crowbar_framework/app/controllers/nodes_controller.rb
+++ b/crowbar_framework/app/controllers/nodes_controller.rb
@@ -105,18 +105,14 @@ class NodesController < ApplicationController
           begin
             dirty = false
             node = NodeObject.find_node_by_name node_name
+            is_allocated = node.allocated?
 
-            unless node.alias == node_attributes["alias"]
-              node.force_alias = node_attributes["alias"]
+            if node_attributes["allocate"] and not is_allocated
+              node.allocate!
               dirty = true
             end
 
-            unless node.public_name == node_attributes["public_name"]
-              node.force_public_name = node_attributes["public_name"]
-              dirty = true
-            end
-
-            unless node.allocated?
+            unless is_allocated
               unless node.target_platform == node_attributes["target_platform"]
                 node.target_platform = node_attributes["target_platform"]
                 dirty = true
@@ -128,8 +124,13 @@ class NodesController < ApplicationController
               end
             end
 
-            if node_attributes["allocate"] and not node.allocated?
-              node.allocate!
+            unless node.alias == node_attributes["alias"]
+              node.force_alias = node_attributes["alias"]
+              dirty = true
+            end
+
+            unless node.public_name == node_attributes["public_name"]
+              node.force_public_name = node_attributes["public_name"]
               dirty = true
             end
 


vuntz:
@tboerger hrm, don't we have the same issue for the target platform and the license key, then?

vuntz:
(or am I misunderstanding that this is caused by `allocated!` that somehow resets some stuff?)

tboerger:
No, because the allocate gets done first, afterwards the other attributes are getting set.

tboerger:
> (or am I misunderstanding that this is caused by allocated! that somehow resets some stuff?)

I think it is related to the save call within the ```allocate!``` method. Eveything assigned after that seems to work.

vuntz:
@tboerger thanks for the clarification. So we also need to move the code setting the license and the target plarform which is before `allocate!` :-)

tboerger:
@vuntz Moved that as well.

crowbar_framework/app/controllers/nodes_controller.rb
```diff
@@ -106,13 +106,8 @@ def bulk
...
-            unless node.alias == node_attributes["alias"]
-              node.force_alias = node_attributes["alias"]
-              dirty = true
-            end
-
-            unless node.public_name == node_attributes["public_name"]
-              node.force_public_name = node_attributes["public_name"]
+            if node_attributes["allocate"] and not node.allocated?
+              node.allocate!
               dirty = true
```
vuntz: You need to save `node.allocated?` before allocating the node, as it's used later on :-)

vuntz:
In general, I'm confused because we changed this in https://github.com/crowbar/barclamp-crowbar/pull/1166

@rhafer any opinion?

crowbar_framework/app/controllers/nodes_controller.rb
tboerger: https://github.com/crowbar/barclamp-crowbar/pull/1240/files#diff-40544af9c69a22d34a1c8859eb0c33efR108

tboerger:
> In general, I'm confused because we changed this in #1166
> 
> @rhafer any opinion?

Yes, but now we are saving the allocated state before the allocate! call.

vuntz:
+1

dirkmueller:
is its possible to write a spec test for this?

tboerger:
Then we need fixtures for unallocated nodes and the allocation have to be mocked... Not really useful IMHO.

bkutil:
There are already node fixtures that you can return from the `NodeObject.find_...` (IIRC they are called 'testing' and 'admin'). Checking for allocation might / might not need to be stubbed. But overall, it should not be that hard.

bkutil:
+1. @tboerger ping

tboerger:
@bkutil Currently I don't really know how to test that exactly and I'm still on the dark side of hawk :)

rsalevsky:
Let's merge this PR due to writing specs for this is atm to complex. I put it on the todo for refactoring the spec's in general.

PR #1242, opened 2015-02-20
Do not use node handle as fallback for switch name (bnc#918785)

vuntz:
This is just wrong: if we don't know the switch info, then let's not
pretend it's something else.

This fixes nodes appearing in different groups by default when there's
no switch info.

https://bugzilla.suse.com/show_bug.cgi?id=918785


bkutil:
+1

vmoravec:
I tested the patch and the node grouping has been fixed by that. There is a single group named `sw-unknown` now on the crowbard dashboard.

tboerger:
:+1:

PR #1243, opened 2015-02-24
Render 404 on barclamp_index in HTML (bnc#916221)

bkutil:
This action is not rendered in HTML, only as JSON when using the API.

Raise 404 when viewed as HTML, so rails will render it's standard
not-found page, as suggested in the BNC entry.

cc @vmoravec

### DIFF
--- a/crowbar_framework/app/controllers/barclamp_controller.rb
+++ b/crowbar_framework/app/controllers/barclamp_controller.rb
@@ -39,7 +39,7 @@ class BarclampController < ApplicationController
   def barclamp_index
     @barclamps = ServiceObject.all
     respond_to do |format|
-      format.html { render :template => 'barclamp/barclamp_index' }
+      format.html { raise ActionController::RoutingError.new('Not Found') }
       format.xml  { render :xml => @barclamps }
       format.json { render :json => @barclamps }
     end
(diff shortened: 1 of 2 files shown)


tboerger:
I think it should be enough to remove the response for html or not?

bkutil:
Hrm, maybe you're right. Let me check what happens then (and if rails rescues that error by rendering the standard page).

bkutil:
Ok, it raises `ActionController::UnknownFormat`, which will bubble up and the user would get 'internal server error' page. So the patch should work, or we could add a rescue_from and return 404/raise NotFound for these errors.

vmoravec:
I tested the patch together with the @tboerger proposal. The original patch works fine. After I removed the html response, I got empty page (no html source) in browser on reload.

vmoravec:
I noticed that crowbar log level is set to DEBUG, is this intentional? I can see `ActiveRecord::SessionStore::Session Load` lines logged every 30 seconds in `/var/log/crowbar/production.log` .

bkutil:
@vmoravec I'm not sure, but maybe @vuntz can tell more?

vuntz:
It is intentional as this log level is required to remove the remaining bugs in crowbar.

That being said, the session load message is annoying. Don't know if we could get rid of it... Probably not easily.

bkutil:
@vuntz @vmoravec can we merge this?

vmoravec:
+1

vmoravec:
There might be some way to resolve the flood of `ActiveRecord::SessionStore::Session` records with `Rails.logger.silence` thing in the `nodes_controller`, but I'm not sure about the implementation.

bkutil:
@vmoravec that could be done. Would you mind creating a trello card for this?

PR #1244, opened 2015-02-24
Increase timeout for nodelist flashes (bnc#917541)

bkutil:
The 10s should be enough for the reader of the flash message and he can
dismiss it sooner, if needed. OTOH, the flash will still disappear
automatically, so it won't take up display space.

We can still decide otherwise and disable the 'auto-hide'.


bkutil:
cc @vmoravec

vmoravec:
10 seconds is much better then 3s.
+1

tboerger:
Ok from my side

PR #1246, opened 2015-02-25
Handle case where device name is linkname

toabctl:
When checking if a disk is already claimed, all links to the claimed
disks are checked as well. Allow the case where the real device name is
given as possible linkname. This can happen when using virtual hardware
where the devicename might be "/dev/vda".

Related to bnc#919469


tboerger:
Sounds reasonable :+1:

PR #1247, opened 2015-02-26
Fix for link_to_device on SLE12 for virtio disks, (bnc#919469)

mjura:
Bug https://bugzilla.suse.com/show_bug.cgi?id=919469

This is fix when virtio disk and the target platform is one that might not have the "by-path" links (e.g. SLES 12). Avoid using "by-path".

This can occur for `{"/dev/vda"=>{"owner"=>"Boot"}}`

### DIFF
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -1351,7 +1351,7 @@ class NodeObject < ChefObject
     return true if device == linkname
     lookup_and_name = linkname.gsub(/^\/dev\/disk\//, '').split(File::SEPARATOR, 2)
     linked_devs = @node[:block_device][device][:disks][lookup_and_name[0]] rescue []
-    linked_devs.include?(lookup_and_name[1])
+    linked_devs.include?(lookup_and_name[1]) rescue false
   end
 
   # IMPORTANT: keep these paths in sync with


crowbar_framework/app/models/node_object.rb
```diff
@@ -1334,6 +1334,22 @@ def boot_device!(device)
...
+      if claimed_name =~ /^\/dev\/vd[a-z]+$/
+        virtio_by_path_platforms = %w(
+          ubuntu-12.04
+          redhat-6.2
+          redhat-6.4
+          centos-6.2
+          centos-6.4
+          suse-11.3
+        )
+        next unless virtio_by_path_platforms.include?(@node[:target_platform])
```
vuntz: I think you don't want to use `next` here, do you?

crowbar_framework/app/models/node_object.rb
```diff
@@ -1334,6 +1334,22 @@ def boot_device!(device)
...
+      end
+
```
toabctl: Hm. That feels wrong. ```unique_name_already_claimed_by(device)``` is just there to compare the given ```device``` with the already ```claimed_disks``` from ```crowbar_wall```. And the code you added here is already in the method ```unique_device_for()```.

rhafer:
Wasn't https://github.com/crowbar/barclamp-crowbar/pull/1246 supposed to fix this already?

crowbar_framework/app/models/node_object.rb
mjura: Yes I do, I want to skip for next interation if `claimed_name == /dev/vd[a-z]` and `target_platform` is different than on list

crowbar_framework/app/models/node_object.rb
mjura: Problem is that `crowbar_wall[:claimed_disks]` will return `{"/dev/vda"=>{"owner"=>"Boot"}}` and on SLE12 there is no `by-path`

mjura:
> Wasn't #1246 supposed to fix this already?

PR #1246 didn't fix this bug

crowbar_framework/app/models/node_object.rb
vuntz: I'd think we want to use `return false` in `find`. But maybe `next` work too :-)

crowbar_framework/app/models/node_object.rb
vuntz: But the ohai data should not have anything for `by-path` either in that case.

rhafer: Maybe from a previous ohai run? When it still was in the SLE11-based sleshammer?

mjura: On SLE12 node, where /dev/vda is root device
```
chef > node[:block_device]['vda'][:disks].inspect
 => "nil" 
```

and comparision for /dev/sda 
````
chef > node[:block_device]['sda'][:disks].inspect
 => "#<Chef::Node::Attribute @normal={...}, @current_normal={}, @default={...}, @current_default={}, @override={...}, @current_override={}, @automatic={...}, @current_automatic={\"by-path\"=>[\"pci-0000:00:06.0-scsi-0:0:0:0\"], \"by-id\"=>[\"scsi-0QEMU_QEMU_HARDDISK_drive-scsi0-0-0\"]}, @current_nesting_level=[:block_device, \"sda\", :disks], @auto_vivifiy_on_read=false, @set_unless_value_present=false, @set_type=nil, @has_been_read=false>"
```

crowbar_framework/app/models/node_object.rb
mjura: As I checked `next`  works better

crowbar_framework/app/models/node_object.rb
rhafer: Ok, but I guess in this case the correct fix would be in link_to_device, no?

crowbar_framework/app/models/node_object.rb
mjura: Now, it is `return false`

crowbar_framework/app/models/node_object.rb
mjura: @rhafer  Thank you, I moved it to `link_to_device`

crowbar_framework/app/models/node_object.rb
```diff
@@ -1349,6 +1349,22 @@ def link_to_device?(device, linkname)
...
+      virtio_by_path_platforms = %w(
+        ubuntu-12.04
+        redhat-6.2
+        redhat-6.4
+        centos-6.2
+        centos-6.4
+        suse-11.3
+      )
+      return false unless virtio_by_path_platforms.include?(@node[:target_platform])
+    end
```
rhafer: > @rhafer Thank you, I moved it to link_to_device

Hm, I guess you missunderstood me a bit. I think this is not the correct fix. IMO the correct fix make sure that ```linked_devs``` cannot be nil a few lines below. (Though I really wonder how that can still happen).

rhafer: Actuall that's a but in https://github.com/crowbar/barclamp-crowbar/pull/1246

toabctl: > Actuall that's a bug in #1246 
> @toabctl is about to fix that.

See https://github.com/crowbar/barclamp-crowbar/pull/1249 for the fix.

vuntz:
So should we close this one?

rhafer:
> So should we close this one?

No. There is yet another issue. @mjura is working on a fix.

mjura:
This is new fix proposal

crowbar_framework/app/models/node_object.rb
```diff
@@ -1351,7 +1351,7 @@ def link_to_device?(device, linkname)
     return true if device == linkname
     lookup_and_name = linkname.gsub(/^\/dev\/disk\//, '').split(File::SEPARATOR, 2)
     linked_devs = @node[:block_device][device][:disks][lookup_and_name[0]] rescue []
-    linked_devs.include?(lookup_and_name[1])
+    linked_devs.include?(lookup_and_name[1]) rescue false
```
vuntz: Our (ab)use of `rescue` is starting to become frightening :/ But I guess I'm fine with this for now :-)

rhafer:
Please update the commit message to explain a bit more why this is needed. And I think this bug is not really limited to virtio. Please also add the bug url. Otherwise I am fine with this.

mjura:
Comment was updated

rhafer:
Thanks. +1

toabctl:
+1

crowbar_framework/app/models/node_object.rb
aspiers: Agreed, this is a bad coding pattern which should be discouraged.

tboerger: And this will bite us heavily when we integrate rubocop :)

aspiers: Glad to hear it! :)

PR #1248, opened 2015-02-26
Fix hideShowClusterConf.js when a cluster is assigned to role on load

vuntz:
We were counting the same cluster twice: once by looking at the
attributes, and once due to the nodeListNodeAllocated event that happens
on page load.


vuntz:
This is a bit fragile, though, because I guess it depends on the order in which the js code is executed...

bkutil:
Actually, this is not that bad. I tested this by trying to change the loading order and was not able to break it. Another check would be good, though.

cc @tboerger

tboerger:
I haven't a strong opinion on that... I'm ok with that change if it makes the process more stable.

PR #1250, opened 2015-02-26
Make NodeObject.group_order return a unique value

vuntz:
The node dashboard relies on the fact that group_order is different for
each node. This was not the case when two nodes are actually VMWare VMs
that are ultimately plugged in the same switch.

So instead of only relying on the information of where the node is
plugged on the switch, also append to that the alias of the node to
ensure group_order is unique.


rhafer:
+1

aspiers:
+1

tboerger:
:+1:

PR #1251, opened 2015-03-09
apply_role: disable 15-minute chef run during apply_role

MaximilianMeister:
partially fixes bsc#857375
https://bugzilla.suse.com/show_bug.cgi?id=857375
https://trello.com/c/zj87VaeB

### DIFF
--- a/crowbar_framework/app/models/chef_object.rb
+++ b/crowbar_framework/app/models/chef_object.rb
@@ -56,4 +56,23 @@ class ChefObject
     file   = Rails.root.join("db", "#{self.chef_type}_#{name}.json")
     File.open(file, "w") { |f| f.write(self.to_json) }
   end
+
+  # Each operating system can have a different path to the init command of a
+  # service.
+  # In order to manipulate services we can tap the provider mapping for the
+  # correct service class to determine the system's correct init command
+  # https://github.com/chef/chef/blob/master/lib/chef/platform/provider_mapping.rb
+  def self.service_command(platform, version, service_name, action)
+    provider_service_class = Chef::Platform.find(platform, version)[:service]
+    case provider_service_class.to_s
+    when "Chef::Provider::Service::Systemd"
+      "systemctl #{action} #{service_name}"
+    when "Chef::Provider::Service::Upstart"
+      "/etc/init.d/#{service_name} #{action}"
+    when "Chef::Provider::Service::Redhat"
+      "service #{service_name} #{action}"
+    else
+      nil
+    end
+  end
 end
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -1268,6 +1268,19 @@ class NodeObject < ChefObject
     @node["crowbar_wall"]["status"]["ipmi"]["address_set"]
   end
 
+  def run_service(service_name, action)
+    platform, version = @node["platform"], @node["platform_version"]
+    case service_name
+    when :chef
+      service_cmd = ChefObject.service_command(platform, version, "chef-client", action)
+    end
+    if service_cmd.nil?
+      Rails.logger.error("run_service: service command for #{platform} #{version} could not be determined")
+    else
+      ssh_cmd(service_cmd)
+    end
+  end
+
   def disk_owner(device)
     if device
       crowbar_wall[:claimed_disks][device][:owner] rescue ""
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1537,7 +1537,7 @@ class ServiceObject
       ssh_cmd = ssh.dup << command
 
       # check if there are currently other chef-client runs on the node
-      wait_for_chef_clients(node)
+      wait_for_chef_clients(node, :logger => false)
       # check if the node is currently rebooting
       wait_for_reboot(node)
 
@@ -1557,15 +1557,21 @@ class ServiceObject
 
     # wait for chef clients on all nodes
     node_list.each do |node_name|
-      wait_for_chef_clients(node_name)
+      wait_for_chef_clients(node_name, :logger => true)
     end if action == :stop
   end
 
   private
 
-  def wait_for_chef_clients(node_name)
-    unless RemoteNode.chef_ready?(node_name, 1200)
-      STDERR.puts "Waiting for already running chef-clients on #{node_name} failed"
+  def wait_for_chef_clients(node_name, options = {})
+    options = if options.fetch(:logger)
+      {:logger => @logger}
+    else
+      {}
+    end
+    @logger.debug("wait_for_chef_clients: Waiting for already running chef-clients on #{node_name}.")
+    unless RemoteNode.chef_ready?(node_name, 1200, 10, options)
+      @logger.error("Waiting for already running chef-clients on #{node_name} failed.")
       exit(1)
     end
   end
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1563,9 +1563,9 @@ class ServiceObject
 
   private
 
-  def wait_for_chef_clients(node)
-    unless RemoteNode.chef_ready?(node, 1200)
-      STDERR.puts "Waiting for already running chef-clients on #{node} failed"
+  def wait_for_chef_clients(node_name)
+    unless RemoteNode.chef_ready?(node_name, 1200)
+      STDERR.puts "Waiting for already running chef-clients on #{node_name} failed"
       exit(1)
     end
   end
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1106,6 +1106,10 @@ class ServiceObject
     end
     new_elements = expanded_new_elements
 
+    # stop chef daemon on all nodes
+    chef_daemon_nodes = new_elements.values.flatten
+    chef_daemon(:stop, chef_daemon_nodes)
+
     # save list of expanded elements, as this is needed when we look at the
     # old role
     if new_elements != new_deployment["elements"]
@@ -1416,6 +1420,9 @@ class ServiceObject
     restore_to_ready(all_nodes)
     process_queue unless in_queue
     [200, {}]
+  ensure
+    # start chef daemon on all nodes
+    chef_daemon(:start, chef_daemon_nodes)
   end
 
   def apply_role_pre_chef_call(old_role, role, all_nodes)
@@ -1539,6 +1546,19 @@ class ServiceObject
       # check if we need to wait for a node reboot
       wait_for_reboot(node)
     }
+  end
+
+  def chef_daemon(action, node_list)
+    node_list.each do |node_name|
+      node = NodeObject.find_node_by_name(node_name)
+      @logger.debug "apply_role: #{action.to_s} chef service on #{node_name}"
+      node.run_service :chef, action
+    end
+
+    # wait for chef clients on all nodes
+    node_list.each do |node_name|
+      wait_for_chef_clients(node_name)
+    end if action == :stop
   end
 
   private


vuntz:
We also need to check that no chef-client is currently running after having stopped the daemon (and wait for all of them if some are running).

aspiers:
Agree with @vuntz.  Thanks for doing this though, will be great to have a partial fix!

MaximilianMeister:
@vuntz do you mean check for a (possible) daemonized chef-client run? the runs that are triggered within apply_role already have a process waiting mechanism afaics

vuntz:
@MaximilianMeister I know, but this check is actually a bit late: in case the chef-client runs on various nodes need to be syncrhonized, having an existing chef-client run will delay one node and cause a failure.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1084,6 +1084,17 @@ def apply_role(role, inst, in_queue)
     element_order = new_deployment["element_order"]
 
     #
+    # stop automatic chef-client on all nodes
+    #
+    nodes_to_freeze_chef_service = new_elements.values.flatten
+    nodes_to_freeze_chef_service.each do |node_name|
+      node = NodeObject.find_node_by_name(node_name)
+      wait_for_chef_clients(node)
```
MaximilianMeister: @vuntz @aspiers do you think that would do it?

MaximilianMeister: I just saw the `wait_for_chef_clients` method has an `exit(1)` after a timeout, would this be acceptable in this case as well?

crowbar_framework/app/models/node_object.rb
```diff
@@ -1249,6 +1249,14 @@ def poweroff
     bmc_cmd("power off")
   end
 
+  def stop_chef_service
+    ssh_cmd("/usr/sbin/rcchef-client stop")
```
tboerger: This will be fun if we add back integration for multiple operating systems :(

aspiers: That's a very good point; this should do an OS check and fail with an `unsupported OS` error if not SUSE.  Assuming we are on SUSE is not consistent with the rest of the Crowbar project.

tboerger: Maybe we can load chef client implementations here to invoke the correct method?

aspiers: Not sure it's possible to execute recipe/LWRP code from within Rails...

tboerger: We are already including the chef client, and i dont want to execute the lwrp directly, i just want to fetch the correct command from the platform mapping :)

aspiers: If it can be done, it's probably a good idea :)

crowbar_framework/app/models/service_object.rb
```diff
@@ -1412,6 +1423,15 @@ def apply_role(role, inst, in_queue)
       return [ 405, message ]
```
vuntz: Hrm. See the `return` here? If we return and don't restart the service, then we lose :-)

crowbar_framework/app/models/service_object.rb
```diff
@@ -1084,6 +1084,17 @@ def apply_role(role, inst, in_queue)
     element_order = new_deployment["element_order"]
 
     #
+    # stop automatic chef-client on all nodes
+    #
+    nodes_to_freeze_chef_service = new_elements.values.flatten
+    nodes_to_freeze_chef_service.each do |node_name|
+      wait_for_chef_clients(node_name)
```
vuntz: You need to wait after stopping the service, not before.

MaximilianMeister: done

crowbar_framework/app/models/service_object.rb
MaximilianMeister: very true. hmm, that is tricky, i counted 6 returns in `apply_role`

MaximilianMeister: solved by `ensure` at the end of `apply_role`
thanks @tboerger for the tip :)

MaximilianMeister:
I addressed the multi os problem, and split the commits with a bit more logic. @tboerger @bkutil @aspiers @vuntz if you have some time please have another look

aspiers:
Getting dangerously close to looking great ;-)  Thanks!

crowbar_framework/app/models/node_object.rb
```diff
@@ -1249,6 +1249,22 @@ def poweroff
...
 
+  def service(service_name, action)
+    case service_name
+    when :chef
+      provider_service_object = ChefObject.get_provider_service_instance(self, "chef-client")
+    end
+
+    case action
+    when :start
+      provider_service_object.start_service
```
MaximilianMeister: @aspiers would be really nice if calling the correct instance method of `Chef::Provider::Service::*` just worked.
each provider service class has the right instance methods to start/stop/etc services.
but instead i get a permission error:
```
Expected process to exit with [0], but received '1'
---- Begin output of /sbin/service chef-client start ----
STDOUT:
STDERR: service: only root can use service
---- End output of /sbin/service chef-client start ----
Ran /sbin/service chef-client start returned 1
```
do you have an idea how to circumvent that?

tboerger: That will never work. This command gets executed on the admin node. You need to execute it on the node itself.

MaximilianMeister:
unfortunately there is not enough information in our metadata to avoid hardcoding the service path to get the right init command in the end. if someone has a better idea, i am running out of them. please have another look.
thanks

crowbar_framework/app/models/chef_object.rb
```diff
@@ -56,4 +56,23 @@ def export(name = nil)
...
+  # https://github.com/chef/chef/blob/master/lib/chef/platform/provider_mapping.rb
+  def self.get_service_init_cmd(platform, version, service_name, action)
+    provider_service_class = Chef::Platform.find(platform, version)[:service]
+    case provider_service_class.to_s
+    when "Chef::Provider::Service::Systemd"
+      "/sbin/systemctl #{action} #{service_name}"
+    when "Chef::Provider::Service::Upstart"
+      "/etc/init.d/#{service_name} #{action}"
+    when "Chef::Provider::Service::Redhat"
+      "/sbin/service #{service_name} #{action}"
```
MaximilianMeister: unfortunately there is not enough information in our metadata to avoid hardcoding the service path to get the right init command in the end. if someone has a better idea, i am running out of them. please have another look.
thanks

crowbar_framework/app/models/node_object.rb
```diff
@@ -1249,6 +1249,14 @@ def poweroff
     bmc_cmd("power off")
   end
 
+  def service(service_name, action)
+    case service_name
+    when :chef
+      init_cmd = ChefObject.get_service_init_cmd(@node["platform"], @node["platform_version"], "chef-client", action)
+    end
+    ssh_cmd(init_cmd) unless init_cmd.nil?
```
aspiers: If `init_cmd` is `nil` then surely there should be at least a warning, if not an error?

aspiers: And what happens on Hyper-V hosts?

MaximilianMeister: > If init_cmd is nil then surely there should be at least a warning, if not an error?

agree

> And what happens on Hyper-V hosts?

i have no idea, but we don't have ssh on them anyways iirc.

MaximilianMeister: > And what happens on Hyper-V hosts?

i think this should be fixed in another place. I am not sure how `apply_role` works for a HyperV node but `run_remote_chef_client` which is executed during `apply_role` goes over ssh as well, so i think this here should not be a problem on the HyperV host.

crowbar_framework/app/models/chef_object.rb
```diff
@@ -56,4 +56,23 @@ def export(name = nil)
     file   = Rails.root.join("db", "#{self.chef_type}_#{name}.json")
     File.open(file, "w") { |f| f.write(self.to_json) }
   end
+
+  # Each operating system can have a different path to the init command of a
+  # service.
+  # In order to manipulate services we can tap the provider mapping for the
+  # correct service class to determine the system's correct init command
+  # https://github.com/chef/chef/blob/master/lib/chef/platform/provider_mapping.rb
+  def self.get_service_init_cmd(platform, version, service_name, action)
```
tboerger: why not more simple like ```service_command```?

tboerger:
LGTM

MaximilianMeister:
just changed method name to `service_command`

tboerger:
:+1:

MaximilianMeister:
@aspiers may I ask you to have another look here? I added a Sprint 19 Milestone, so this time it's serious :-) Thanks

crowbar_framework/app/models/node_object.rb
```diff
@@ -1249,6 +1249,19 @@ def poweroff
...
+    when :chef
+      service_cmd = ChefObject.service_command(platform, version, "chef-client", action)
+    end
+    if service_cmd.nil?
+      Rails.logger.error("service: init command for #{platform} #{version} could not be determined")
+    else
+      ssh_cmd(service_cmd)
+    end
+  end
+
```
vuntz: Can you move this a few lines later (before `disk_owner`)? Right now, it's in the middle of some BMC-related methods, which is weird.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1084,6 +1084,17 @@ def apply_role(role, inst, in_queue)
...
     #
+    # stop chef daemon on all nodes
+    #
+    nodes_to_freeze_chef_service = new_elements.values.flatten
+    nodes_to_freeze_chef_service.each do |node_name|
+      node = NodeObject.find_node_by_name(node_name)
+      @logger.debug "stopping chef service on #{node_name} before apply_role"
+      node.service :chef, :stop
+      wait_for_chef_clients(node_name)
+    end
```
vuntz: A few things:
* this is too early (a few lines below, we might put the apply in the queue)
* this is not taking clusters in consideration (you have `expanded_new_elements` for that a few lines later)

crowbar_framework/app/models/service_object.rb
```diff
@@ -1543,9 +1561,10 @@ def run_remote_chef_client(node, command, logfile_name)
 
   private
 
-  def wait_for_chef_clients(node)
-    unless RemoteNode.chef_ready?(node, 1200)
-      STDERR.puts "Waiting for already running chef-clients on #{node} failed"
+  def wait_for_chef_clients(node_name)
+    @logger.debug("wait_for_chef_clients: Waiting for already running chef-clients on #{node_name}.")
```
vuntz: Hrm, does this work fine from the other place where this is being called (`run_remote_chef_client`)? (it's from a fork, so I'm not sure `@logger.debug` worls there)

Actually, with this change, we can probably remove that other `wait_for_chef_clients` call, can't we?

crowbar_framework/app/models/service_object.rb
```diff
@@ -1084,6 +1084,17 @@ def apply_role(role, inst, in_queue)
...
 
     #
+    # stop chef daemon on all nodes
+    #
+    nodes_to_freeze_chef_service = new_elements.values.flatten
+    nodes_to_freeze_chef_service.each do |node_name|
+      node = NodeObject.find_node_by_name(node_name)
+      @logger.debug "stopping chef service on #{node_name} before apply_role"
+      node.service :chef, :stop
+      wait_for_chef_clients(node_name)
```
vuntz: It's best to do two loops: one where we stop chef-client everywhere, and then one where we wait. Right now, you might hit this:
* stop on node1, see that it's running on node1 and then wait 5 minutes for it to complete
* stop on node2, and same thing, wait 5 minutes
* etc.

If you do two loops, you won't wait 5+5+5+5+5+... minutes but max(5,5,5,...) minutes.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1249,6 +1249,19 @@ def poweroff
     bmc_cmd("power off")
   end
 
+  def service(service_name, action)
```
vuntz: While this method name makes sense in the context of this pull request, I'm not convinced the name makes it obvious what this is about outside of this context. Maybe `run_service`?

crowbar_framework/app/models/node_object.rb
```diff
@@ -1249,6 +1249,19 @@ def poweroff
...
+    platform, version = @node["platform"], @node["platform_version"]
+    case service_name
+    when :chef
+      service_cmd = ChefObject.service_command(platform, version, "chef-client", action)
+    end
+    if service_cmd.nil?
+      Rails.logger.error("service: init command for #{platform} #{version} could not be determined")
```
vuntz: nitpick: the message is about an `init command` but everything else we do is about a `service command` :-)

crowbar_framework/app/models/chef_object.rb
vuntz: Do we actually need the full path? I'd expect `service` to be enough.

crowbar_framework/app/models/chef_object.rb
```diff
@@ -56,4 +56,23 @@ def export(name = nil)
...
+  # Each operating system can have a different path to the init command of a
+  # service.
+  # In order to manipulate services we can tap the provider mapping for the
+  # correct service class to determine the system's correct init command
+  # https://github.com/chef/chef/blob/master/lib/chef/platform/provider_mapping.rb
+  def self.service_command(platform, version, service_name, action)
+    provider_service_class = Chef::Platform.find(platform, version)[:service]
+    case provider_service_class.to_s
+    when "Chef::Provider::Service::Systemd"
+      "/sbin/systemctl #{action} #{service_name}"
```
vuntz: FWIW:
```
# which systemctl
/usr/bin/systemctl
```

crowbar_framework/app/models/service_object.rb
MaximilianMeister: > Hrm, does this work fine from the other place where this is being called (run_remote_chef_client)? (it's from a fork, so I'm not sure @logger.debug worls there)

Good point, I have no clue I would need to test that. You have any suggestions?

> Actually, with this change, we can probably remove that other wait_for_chef_clients call, can't we?

Not sure if I understand. which other `wait_for_chef_clients` call are you referring to?

crowbar_framework/app/models/node_object.rb
MaximilianMeister: done

crowbar_framework/app/models/service_object.rb
```diff
@@ -1106,6 +1106,19 @@ def apply_role(role, inst, in_queue)
...
+    nodes_to_freeze_chef_service.each do |node_name|
+      node = NodeObject.find_node_by_name(node_name)
+      @logger.debug "stopping chef service on #{node_name} before apply_role"
+      node.run_service :chef, :stop
+    end
+
+    # wait for chef clients on all nodes
+    nodes_to_freeze_chef_service.each do |node_name|
+      wait_for_chef_clients(node_name)
+    end
```
MaximilianMeister: @vuntz is this what you had in mind with the second loop?

crowbar_framework/app/models/service_object.rb
vuntz: > Not sure if I understand. which other `wait_for_chef_clients` call are you referring to?

The one from inside the fork, see https://github.com/crowbar/barclamp-crowbar/blob/master/crowbar_framework/app/models/service_object.rb#L1533

And if we drop it, then we can ignore the issue about `@logger` and we can also replace `STDERR.puts` with proper log handling.

crowbar_framework/app/models/service_object.rb
vuntz: I believe so, yes. Thanks!

crowbar_framework/app/models/chef_object.rb
```diff
@@ -56,4 +56,23 @@ def export(name = nil)
...
+  # https://github.com/chef/chef/blob/master/lib/chef/platform/provider_mapping.rb
+  def self.service_command(platform, version, service_name, action)
+    provider_service_class = Chef::Platform.find(platform, version)[:service]
+    case provider_service_class.to_s
+    when "Chef::Provider::Service::Systemd"
+      "/user/bin/systemctl #{action} #{service_name}"
+    when "Chef::Provider::Service::Upstart"
+      "/etc/init.d/#{service_name} #{action}"
+    when "Chef::Provider::Service::Redhat"
+      "/sbin/service #{service_name} #{action}"
```
vuntz: So for the record, I'm still not convinced we need the full path to `service` and `systemctl`, but I don't mind for now :-)

vuntz:
Thanks for the quick update! This looks mostly good to me; there's still the bit at https://github.com/crowbar/barclamp-crowbar/pull/1251#discussion_r29241948 to investigate, and then I think it'll be ready!

crowbar_framework/app/models/service_object.rb
MaximilianMeister: @vuntz thanks for the review. i will drop it and give it another test run. will ping you again later

MaximilianMeister:
@vuntz I dropped `wait_for_chef_client` from `run_remote_chef_client` now. see https://github.com/MaximilianMeister/barclamp-crowbar/commit/e197e599bbf162dd75b8df23a68a948804fba35d

tested again, works

crowbar_framework/app/models/service_object.rb
```diff
@@ -1543,9 +1561,10 @@ def run_remote_chef_client(node, command, logfile_name)
 
   private
 
-  def wait_for_chef_clients(node)
-    unless RemoteNode.chef_ready?(node, 1200)
-      STDERR.puts "Waiting for already running chef-clients on #{node} failed"
+  def wait_for_chef_clients(node_name)
+    @logger.debug("wait_for_chef_clients: Waiting for already running chef-clients on #{node_name}.")
+    unless RemoteNode.chef_ready?(node_name, 1200)
+      STDERR.puts "Waiting for already running chef-clients on #{node_name} failed"
```
rsalevsky: Can we use here also a logger error handling?

MaximilianMeister: ups, @vuntz mentioned that already, slipped through... thanks for the hint

crowbar_framework/app/models/service_object.rb
```diff
@@ -1543,9 +1561,10 @@ def run_remote_chef_client(node, command, logfile_name)
 
   private
 
-  def wait_for_chef_clients(node)
-    unless RemoteNode.chef_ready?(node, 1200)
-      STDERR.puts "Waiting for already running chef-clients on #{node} failed"
+  def wait_for_chef_clients(node_name)
+    @logger.debug("wait_for_chef_clients: Waiting for already running chef-clients on #{node_name}.")
+    unless RemoteNode.chef_ready?(node_name, 1200)
```
vuntz: Hrm, I just realized we opened a can of worms: `RemoteNode.chef_ready?` is using `puts` internally :/

Not sure how to best handle that now... @bkutil might have a good idea?

MaximilianMeister: @vuntz Are you confident about the other patches related to the fix?

I fear the worms. As this is really not related to the fix of [857375](https://bugzilla.suse.com/show_bug.cgi?id=857375), can we just drop the logging improvement commit from here and reopen another PR for that? Would be good to get the other stuff merged...

vuntz: The problem is not the logging improvement, the problem is that you're re-using `wait_for_chef_clients` that is really written with the idea that it's going to be called from a `fork`, hence its use of `puts`. And you're using it from inside the rails app instead of from a `fork`.

So the patch itself is not bad; it's just built on something that was not intended for this use.

MaximilianMeister: @vuntz could we use the [chef_clients_running?](https://github.com/crowbar/barclamp-crowbar/blob/master/crowbar_framework/lib/remote_node.rb#L54) method from `lib/remote_node.rb` instead? or better [chef_ready?](https://github.com/crowbar/barclamp-crowbar/blob/master/crowbar_framework/lib/remote_node.rb#L41) then we have `chef_clients_running?` in a loop

MaximilianMeister: ... well I still don't see why this should not work, or why it should be specific for being called inside a fork

vuntz: I probably wasn't clear enough: the code will work, the only issue is the logging. Because the code in `RemoteNode` is designed to be called from a fork, it's using `puts` for logging. Now, with the changes you're doing here, we would also use `RemoteNode` without a fork, and in that case, it's not appropriate to use `puts` for logging. It works, but it's not appropriate.

MaximilianMeister: @vuntz would https://github.com/crowbar/barclamp-crowbar/issues/1265 be good enough to resolve the logging issue here?

crowbar_framework/app/models/chef_object.rb
MaximilianMeister: @vuntz sorry here, I have overlooked your two comments, i guess i force pushed sth here before reading them... :-| anyways I made the changes now. @aspiers thanks for the hint

aspiers: @MaximilianMeister If you don't receive updates to PRs as email notifications, I strongly recommend you enable that :)  It's pretty hard to miss comments if you read them via email.  In this case you might also find http://blog.adamspiers.org/2014/10/05/managing-your-github-notifications-inbox-with-mutt/ useful.

MaximilianMeister:
@vuntz after https://github.com/crowbar/barclamp-crowbar/pull/1265 is merged, the issue with the loggin here should be resolved

crowbar_framework/app/models/service_object.rb
```diff
@@ -1529,8 +1536,6 @@ def run_remote_chef_client(node, command, logfile_name)
       ssh = ["sudo", "-u", "root", "--", "ssh", "-o", "TCPKeepAlive=no", "-o", "ServerAliveInterval=15", "root@#{node}"]
       ssh_cmd = ssh.dup << command
 
-      # check if there are currently other chef-client runs on the node
-      wait_for_chef_clients(node)
```
vuntz: Thinking out loud: I wonder if we should keep this here, just in case there's a manual chef-client (or the chef-client daemon got restarted). Of course, this means we should change `wait_for_chef_clients` to accept running with and without logger.

What do you think?

(other than that, I'm +1)

MaximilianMeister: you are probably right. I could test it and see if it fails, but I would rather keep it here as well for being on the safe side. will update the PR

crowbar_framework/app/models/service_object.rb
```diff
@@ -1541,11 +1548,30 @@ def run_remote_chef_client(node, command, logfile_name)
...
+    end if action == :stop
+  end
+
   private
 
-  def wait_for_chef_clients(node)
-    unless RemoteNode.chef_ready?(node, 1200)
-      STDERR.puts "Waiting for already running chef-clients on #{node} failed"
+  def wait_for_chef_clients(node_name, options = {})
+    logger = if options.fetch(:logger)
```
vuntz: Maybe name that variable `options`.

vuntz:
Minor nitpick; other than that, I'm +1 (assuming this has been tested).

Thanks a lot for all the changes you did based on feedback!

MaximilianMeister:
@vuntz changed the variable name, tested one more time, works as expected :) even the logs are there ;)

vuntz:
+1

tboerger:
LGTM

MaximilianMeister:
@aspiers you think this is ready to merge?

rsalevsky:
I would say lets merge it. :+1:

MaximilianMeister:
@aspiers ping :)

jdsn:
@aspiers Please help team green to finish this PR and its trello card. Please have a final review of this PR and if possible merge it. Thanks.

tboerger:
Talked about it in our meeting, we are qaiting far too long for reviews and we already have multiple +1, merging.

aspiers:
Sorry, I only just saw these requests.  I had a long backlog from Vancouver etc.  Feel free to ping me on IRC etc. in future if I don't respond quick enough.

MaximilianMeister:
@aspiers no problem, hope you're fine with the PR as it is

> Feel free to ping me on IRC etc. in future if I don't respond quick enough.

Just for the record ;-)
```
May 27 10:59:31 <mmeister> aspiers: could you please have a (final) look at https://github.com/crowbar/barclamp-crowbar/pull/1251? since you reported the bug. Thanks
```

aspiers:
Yeah, well I didn't say it would be enough to ping me once ;-)  I think the PR is fine although I still suspect that `#service_command` class method belongs somewhere else.

PR #1254, opened 2015-03-19
[WIP] feature/sqlite storage

bkutil:
This is a wip branch attempting to move proposals out of couch db into sqlite.

cc: @vuntz @tboerger @aspiers

Re-creating the pull request against a feature branch as this is not mergeable into master ATM.

The idea behind this is to create a `ProposalObject`-compatible replacement, then keep it in sync by adding hooks into `ProposalObject` `save` and `destroy` methods.

Bi-directional sync is still WIP, but it'd allow us to gradually replace the old `ProposalObject`, instead of making the switch in one step.

To reduce duplication, all common methods are refactored out into `ProposalMethods` module, then included into both classes. The rest is implemented by each class separately.

The API compatibility is tested using rspec, by checking that  both objects respond to the same set of public methods (and methods in the `ProposalMethods` module), and that getters (0-arity methods) in fact return the same result. Data from fixtures (in the case of `ProposalObject`) and those stored in sqlite (for `Proposal` model) are compared.

The most notable difference between the interfaces is the `id` method, which is used by `ActiveRecord` and so it cannot be implemented by `Proposal` (`key` is the corresponding method). There aren't many callers of `id`, so it should not be a big deal anyway.

### DIFF
--- a/crowbar_framework/app/models/chef_object.rb
+++ b/crowbar_framework/app/models/chef_object.rb
@@ -51,9 +51,9 @@ class ChefObject
     str.gsub("-:") { |c| '\\' + c }
   end
 
-  def export(name = nil)
-    name ||= self.respond_to?(:name) ? self.name : "unknown"
-    file   = Rails.root.join("db", "#{self.chef_type}_#{name}.json")
-    File.open(file, "w") { |f| f.write(self.to_json) }
+  def export(name = nil, obj = self)
+    name ||= obj.respond_to?(:name) ? obj.name : "unknown"
+    file   = Rails.root.join("db", "#{obj.chef_type}_#{name}.json")
+    File.open(file, "w") { |f| f.write(obj.to_json) }
   end
 end
--- a/crowbar_framework/app/models/chef_object.rb
+++ b/crowbar_framework/app/models/chef_object.rb
@@ -55,9 +55,13 @@ class ChefObject
   # can be exported in a same way as the ProposalObject. When the replacement
   # is completed, remove it and implement the export in the Proposal.
   # Also check the logging barclamp that it did not break.
-  def export(name = nil, obj = self)
+  def self.export(obj, name = nil)
     name ||= obj.respond_to?(:name) ? obj.name : "unknown"
     file   = Rails.root.join("db", "#{obj.chef_type}_#{name}.json")
     File.open(file, "w") { |f| f.write(obj.to_json) }
+  end
+
+  def export(name = nil)
+    self.class.export(self, name)
   end
 end
--- a/crowbar_framework/app/models/chef_object.rb
+++ b/crowbar_framework/app/models/chef_object.rb
@@ -51,6 +51,10 @@ class ChefObject
     str.gsub("-:") { |c| '\\' + c }
   end
 
+  # FIXME: the second argument was added so that the Proposal model
+  # can be exported in a same way as the ProposalObject. When the replacement
+  # is completed, remove it and implement the export in the Proposal.
+  # Also check the logging barclamp that it did not break.
   def export(name = nil, obj = self)
     name ||= obj.respond_to?(:name) ? obj.name : "unknown"
     file   = Rails.root.join("db", "#{obj.chef_type}_#{name}.json")
--- a/crowbar_framework/app/models/proposal.rb
+++ b/crowbar_framework/app/models/proposal.rb
@@ -29,6 +29,30 @@ class Proposal < ActiveRecord::Base
     super
   end
 
+  def self.find_proposals(barclamp)
+    where(:barclamp => barclamp)
+  end
+
+  def self.find_barclamp(barclamp)
+    self.new(:barclamp => barclamp, :name => "template")
+  end
+
+  def self.find_proposal(barclamp, name)
+    where(:barclamp => barclamp, :name => name).first
+  end
+
+  # XXX: the networks will still be backed by ProposalObject for now,
+  # so it is not neccessary to handle them here.
+  # We still need to handle lookups for templates, though.
+  def self.find_proposal_by_id(id)
+    _, barclamp_or_template, name = *id.split("-")
+    if barclamp_or_template == "template"
+      self.new(:barclamp => name, :name => "template")
+    else
+      where(:barclamp => barclamp_or_template, :name => name).first
+    end
+  end
+
   def to_json
     self.properties.to_json
   end
--- a/crowbar_framework/app/models/proposal.rb
+++ b/crowbar_framework/app/models/proposal.rb
@@ -17,7 +17,7 @@ class Proposal < ActiveRecord::Base
   validate  :name, :name_not_on_blacklist
   validates :name, uniqueness: { scope: :barclamp }
 
-  after_initialize :load_properties_template
+  after_initialize :load_properties_template, :set_default_name
   before_save      :update_proposal_id
 
   # XXX: a 'registered' barclamp could have a has_many :proposals and have a factory
@@ -110,6 +110,10 @@ class Proposal < ActiveRecord::Base
       # XXX: this assumes barclamps are cloned in the same directory
       Rails.root.join("../../barclamp-#{self.barclamp}/chef/data_bags/crowbar/bc-template-#{self.barclamp}.json").expand_path
     end
+  end
+
+  def set_default_name
+    self.name ||= "template"
   end
 
   def update_proposal_id
--- a/crowbar_framework/app/models/proposal.rb
+++ b/crowbar_framework/app/models/proposal.rb
@@ -67,7 +67,7 @@ class Proposal < ActiveRecord::Base
   end
 
   def export
-    ChefObject.new.export(self.name, self)
+    ChefObject.export(self)
   end
 
   # FIXME: this is not correct, the item of ProposalObject returns
--- a/crowbar_framework/app/models/proposal.rb
+++ b/crowbar_framework/app/models/proposal.rb
@@ -1,3 +1,5 @@
 class Proposal < ActiveRecord::Base
   serialize :properties, JSON
+
+  validates :name, :barclamp, :properties, :presence => true
 end
--- a/crowbar_framework/app/models/proposal.rb
+++ b/crowbar_framework/app/models/proposal.rb
@@ -11,6 +11,7 @@ class Proposal < ActiveRecord::Base
   validates :name, :uniqueness => { :scope => :barclamp }
 
   after_initialize :load_properties_template
+  before_save      :update_proposal_id
 
   # XXX: a 'registered' barclamp could have a has_many :proposals and have a factory
   # method for creating them. Then the check for barclamp arg would not be needed,
@@ -41,5 +42,9 @@ class Proposal < ActiveRecord::Base
 
   def properties_template_path
     Rails.root.join("../../barclamp-#{self.barclamp}/chef/data_bags/crowbar/bc-template-#{self.barclamp}.json").expand_path
+  end
+
+  def update_proposal_id
+    self.properties["id"] = "bc-#{self.barclamp}-#{self.name}"
   end
 end
--- a/crowbar_framework/app/models/proposal.rb
+++ b/crowbar_framework/app/models/proposal.rb
@@ -1,0 +1,3 @@
+class Proposal < ActiveRecord::Base
+  serialize :properties, JSON
+end
--- a/crowbar_framework/app/models/proposal.rb
+++ b/crowbar_framework/app/models/proposal.rb
@@ -3,6 +3,7 @@ class Proposal < ActiveRecord::Base
 
   validates :name, :barclamp, :properties, :presence => true
   validate  :name, :name_not_on_blacklist
+  validates :name, :uniqueness => { :scope => :barclamp }
 
   private
 
--- a/crowbar_framework/app/models/proposal.rb
+++ b/crowbar_framework/app/models/proposal.rb
@@ -2,4 +2,15 @@ class Proposal < ActiveRecord::Base
   serialize :properties, JSON
 
   validates :name, :barclamp, :properties, :presence => true
+  validate  :name, :name_not_on_blacklist
+
+  private
+
+  def name_not_on_blacklist
+    forbidden_names = ["template", "nodes", "commit", "status"]
+
+    if forbidden_names.include?(self.name)
+      self.errors.add(:name, "Name cannot be any of #{forbidden_names.to_sentence}.")
+    end
+  end
 end
--- a/crowbar_framework/app/models/proposal.rb
+++ b/crowbar_framework/app/models/proposal.rb
@@ -70,6 +70,9 @@ class Proposal < ActiveRecord::Base
     ChefObject.new.export(self.name, self)
   end
 
+  # FIXME: this is not correct, the item of ProposalObject returns
+  # couchdb/serialization related attributes. This is equivalent to raw_data
+  # instead.
   def item
     self.properties
   end
--- a/crowbar_framework/app/models/proposal.rb
+++ b/crowbar_framework/app/models/proposal.rb
@@ -1,7 +1,7 @@
 class Proposal < ActiveRecord::Base
   include Crowbar::ProposalMethods
 
-  # FIXME: remove this
+  # FIXME: remove this when the export is properly implemented
   class_attribute :chef_type
 
   self.chef_type = "data_bag_item"
--- a/crowbar_framework/app/models/proposal.rb
+++ b/crowbar_framework/app/models/proposal.rb
@@ -104,7 +104,12 @@ class Proposal < ActiveRecord::Base
   end
 
   def properties_template_path
-    Rails.root.join("../../barclamp-#{self.barclamp}/chef/data_bags/crowbar/bc-template-#{self.barclamp}.json").expand_path
+    if Rails.env.production?
+      Rails.root.join("../chef/data_bags/crowbar/bc-template-#{self.barclamp}.json").expand_path
+    else
+      # XXX: this assumes barclamps are cloned in the same directory
+      Rails.root.join("../../barclamp-#{self.barclamp}/chef/data_bags/crowbar/bc-template-#{self.barclamp}.json").expand_path
+    end
   end
 
   def update_proposal_id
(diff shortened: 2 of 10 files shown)


aspiers:
This is a huge PR so will take time to review - perhaps each team can add (noise) stories to review?

bkutil:
@aspiers I might squash some commits, or re-group the chunks into fewer, better orgranized patches. would that help?

MaximilianMeister:
@bkutil IMO grouping and maybe some squashing sounds good, one thing could be separating the spec's into own commits, so we can review the changes in functionality first, and once it is fine we can review the tests.
or tdd way, turn it around and review the spec's before. i don't know, what did you write first? ;)
thats just my 2 cents, for me this is a bit difficult to review now as it is, due to the amount of changes and due to lack of expertise

aspiers:
Don't squash for the sake of squashing, e.g. squashing 2 unrelated commits together will reduce the number of commits in the PR by 1 but makes the task of reviewing even harder, which is totally pointless.  However in contrast, anything which fixes errors introduced by earlier commits within the same PR *must* be squashed.  Please see https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes and in particular the golden rule:

> If a code change can be split into a sequence of patches/commits, then it should be split. Less is not more. More is more.

Another golden rule is that every commit must result in a 100% test pass.  So you can only add new specs in a spec if they pass (i.e. represent existing functionality).  If a commit adds or changes behaviour, then the corresponding changes/additions to specs must exist within the same commit.

I can't recommend that wiki page highly enough though - it's so incredibly right that I'm going to violate DRY and paste the link a second time ;-)  https://wiki.openstack.org/wiki/GitCommitMessages

crowbar_framework/lib/crowbar/proposal_methods.rb
```diff
@@ -0,0 +1,145 @@
...
+        return "hold" if !bc.has_key? "crowbar-queued" and !bc.has_key? "crowbar-committing"
+        if !bc.key? "crowbar-status" or bc["crowbar-status"] === "success"
+          "ready"
+        else
+          "failed"
+        end
+      end
+    end
+    
+    # nil if not appliciable, true = if success, false if failed
```
vuntz: Typo: applicable

crowbar_framework/lib/crowbar/proposal_methods.rb
```diff
@@ -0,0 +1,145 @@
...
+    def failed?
+       status === 'failed'
+    end
+
+    # for locationlization, will lookup text before the :
```
vuntz: Typo (I guess): localization

crowbar_framework/app/models/chef_object.rb
```diff
@@ -51,9 +51,9 @@ def self.chef_escape(str)
...
   end
 
-  def export(name = nil)
-    name ||= self.respond_to?(:name) ? self.name : "unknown"
-    file   = Rails.root.join("db", "#{self.chef_type}_#{name}.json")
-    File.open(file, "w") { |f| f.write(self.to_json) }
+  def export(name = nil, obj = self)
+    name ||= obj.respond_to?(:name) ? obj.name : "unknown"
+    file   = Rails.root.join("db", "#{obj.chef_type}_#{name}.json")
+    File.open(file, "w") { |f| f.write(obj.to_json) }
```
vuntz: I must admit I find it a bit hacky to abuse ChefObject.export for exporting the new Proposal object. But I understand it might be required for now (as fixing this likely requires changing the logging barclamp).

I see a FIXME at https://github.com/crowbar/barclamp-crowbar/pull/1254/files#diff-7c1b3ff9bea4e2068fbd92ad5b03dcd4R4 for this, but I'd welcome another FIXME here (and possibly a new task in the EPIC for the rearchitecture?)

crowbar_framework/lib/crowbar/proposal_methods.rb
```diff
@@ -0,0 +1,145 @@
+module Crowbar
+  module ProposalMethods
```
vuntz: Just so I understand: is this temporary until we get rid of ProposalObject, or do we plan to keep the public methods split forever?

crowbar_framework/db/migrate/20150312081459_create_proposals.rb
```diff
@@ -0,0 +1,11 @@
+class CreateProposals < ActiveRecord::Migration
+  def change
+    create_table :proposals do |t|
+      t.string :barclamp, :null => false
+      t.string :name, :null => false
+      t.text :properties
```
vuntz: Wondering out loud: I wonder if the attributes and deployment parts of the json would be something we should have split here: that is two text columns, one for attributes and one for deployment.

It might make sense as, longer term, most of what is in the deployment shouldn't actually be in the DB imho (only the role assignment). The rest of the deployment bits should actually be a straight copy from the template.

vuntz: @aspiers I guess you might have an opinion on that.

vuntz: Btw, if we only keep the role assignment in the db, then maybe this needs to be a new table.

(Of course, all this requires quite some work in the other parts of the core to work, so this might be a later step in the branch)

crowbar_framework/lib/crowbar/proposal_methods.rb
```diff
@@ -0,0 +1,145 @@
...
+        )
+      )
+    end
+
+    def category
+      @category ||= BarclampCatalog.category(barclamp)
+    end
+
+    def item
+      @item
```
vuntz: I don't see `@item` being defined for Proposal. Am I missing something?

vuntz: Ah, the method is overriden in Proposal.

crowbar_framework/app/models/proposal_object.rb
```diff
@@ -155,106 +121,36 @@ def barclamp
...
-  def []=(attrib, value)
-    @item[attrib] = value
-  end
+  def save_proposal_in_sqlite
+    attrs = { barclamp: self.barclamp, name: self.name }
 
-  def initialize(x)
-    @item = x
+    prop = Proposal.where(attrs).first_or_initialize(attrs)
+    prop.update(attrs.merge(properties: @item.raw_data)) if @item.raw_data != prop.raw_data
```
vuntz: Will the `@item.raw_data != prop.raw_data` compare strings, or JSON/Hashes? If the former, then I think it's not correct as the string can change without the JSON changing.

crowbar_framework/app/models/proposal.rb
```diff
@@ -0,0 +1,118 @@
+class Proposal < ActiveRecord::Base
+  include Crowbar::ProposalMethods
+
+  # FIXME: remove this
+  class_attribute :chef_type
+
+  self.chef_type = "data_bag_item"
+
+  # FIXME: add proper i18n to errors
```
vuntz: Yes, please :-) We should do that everywhere :-)

crowbar_framework/app/models/proposal.rb
```diff
@@ -0,0 +1,118 @@
...
+
+    super
+  end
+
+  def self.find_proposals(barclamp)
+    where(:barclamp => barclamp)
+  end
+
+  def self.find_barclamp(barclamp)
+    self.new(:barclamp => barclamp, :name => "template")
```
vuntz: Does that mean that we put the template in the database? Do we really need to do it?

crowbar_framework/app/models/proposal.rb
```diff
@@ -0,0 +1,118 @@
...
+    self.properties.to_json
+  end
+
+  # FIXME: equivalent to ProposalObject.id
+  def key
+    if name == "template"
+      "bc-#{self.name}-#{self.barclamp}"
+    else
+      "bc-#{self.barclamp}-#{self.name}"
+    end
```
vuntz: Probably not something for now, but we should definitely have an item in the EPIC tasklist for getting rid of these stupid IDs. The `bc-` prefix is not needed.

crowbar_framework/app/models/proposal.rb
```diff
@@ -0,0 +1,118 @@
...
+  end
+
+  def export
+    ChefObject.new.export(self.name, self)
+  end
+
+  # FIXME: this is not correct, the item of ProposalObject returns
+  # couchdb/serialization related attributes. This is equivalent to raw_data
+  # instead.
```
vuntz: What's the impact of this?

crowbar_framework/app/models/proposal.rb
```diff
@@ -0,0 +1,118 @@
...
+  end
+
+  def raw_data=(value)
+    self.properties = value
+  end
+
+  private
+
+  def name_not_on_blacklist
+    forbidden_names = ["template", "nodes", "commit", "status"]
```
vuntz: Stupid question, but why these forbidden names? I understand template, although I don't necessarily agree that the template should be in the db, but not the others.

crowbar_framework/lib/crowbar/proposal_methods.rb
bkutil: > Just so I understand: is this temporary until we get rid of ProposalObject, or do we plan to keep the public methods split forever?

This is primarily to keep the API's as close as possible in terms of
functionality.

When we get rid of ProposalObject, it'd be good to give this some more
thought as many of these methods are used only in the crowbar view and
could be moved to a 'decorator'. The rest would then go to the Proposal.

crowbar_framework/db/migrate/20150312081459_create_proposals.rb
bkutil: > Wondering out loud: I wonder if the attributes and deployment parts of the json would be something we should have split here: that is two text columns, one for attributes and one for deployment.
> 
> It might make sense as, longer term, most of what is in the deployment shouldn't actually be in the DB imho (only the role assignment). The rest of the deployment bits should actually be a straight copy from the template.
> 

Hmm, I'm not sure I see extra advantages of having the deployment and
attributes separated - they're a part of single template file. Do you
have any examples, or is this more of a 'this is better' hunch? :)

Maybe when the deployment part of the template would change and we'd
have separate templates for attributes and deployment, then
we wouldn't have to migrate the db that often.

For now, it keeps the existing and the new implementation similar
and we can do the split when the migration is over.

Should I create a research card for this?

crowbar_framework/app/models/proposal.rb
bkutil: > Yes, please :-) We should do that everywhere :-)

I have that patch ready, I just did not want to add more patches until
the current ones are reviewed :)

crowbar_framework/app/models/proposal.rb
bkutil: > > +  def self.find_barclamp(barclamp)
> > +    self.new(:barclamp => barclamp, :name => "template")
> 
> Does that mean that we put the template in the database? Do we really need to do it?

This just creates a new object, but does not save it (until the caller
does that, of course).

crowbar_framework/app/models/proposal.rb
bkutil: > > +  # FIXME: this is not correct, the item of ProposalObject returns
> > +  # couchdb/serialization related attributes. This is equivalent to raw_data
> > +  # instead.
> 
> What's the impact of this?

As we AFAIK don't manipulate any of the couchdb metadata directly, this
should not cause any issues.

crowbar_framework/app/models/chef_object.rb
aspiers: Yeah, this looks wrong to me too.  How about a class method which exports any Chef object, and then the instance method wraps around it?

crowbar_framework/app/models/proposal.rb
aspiers: Just checking: you mean `bc-` becomes redundant when the objects are placed in a dedicated table in a real database?  That would make sense to me.

crowbar_framework/app/models/proposal.rb
bkutil: > Stupid question, but why these forbidden names?

I added these validations more as a check that we can do whatever is in
the service object.

The 'forbidden' proposal names are here:

https://github.com/crowbar/barclamp-crowbar/blob/master/crowbar_framework/app/models/service_object.rb#L28

Added in 0bcd8fa206955be2d24113265cc242952ba515f0, by Victor, in
response to an issue in dell bug tracker.

> I understand template, although I don't necessarily agree that the template should be in the db, but not the others.

This validation will prevent anything that has 'template' as a name to
get saved. I can make this a default for Proposal if name is not
specified, so changing the name will be required before saving the
Proposal.

crowbar_framework/spec/models/proposal_object_spec.rb
```diff
@@ -18,6 +18,61 @@
 require 'spec_helper'
 
 describe ProposalObject do
+
+  describe "sqlite sync" do
```
aspiers: OMG, more specs! Extremely cool :satisfied:

crowbar_framework/db/migrate/20150312081459_create_proposals.rb
vuntz: @bkutil so it's a hunch, but based on several things:

* most bits in the deployment part of the proposal should actually never be exposed to the user (nor should the user be allowed to change them, imho)
* the revisions and crowbar-applied bits that are in the deployment part should likely just be straight columns in the database anyway
* having the role assignments available in a different table and in an easy way to query is something that, I guess, will be useful longer term (for the orchestration)
* the other bits should really be static

That being said, I'd definitely agree with doing that as a later step; so a research card is fine with me.

crowbar_framework/app/models/proposal.rb
vuntz: Yeah.

crowbar_framework/app/models/proposal_object.rb
bkutil: > Will the `@item.raw_data != prop.raw_data` compare strings, or JSON/Hashes? If the former, then I think it's not correct as the string can change without the JSON changing.

It should compare hashes, I covered this by a test here:

https://github.com/crowbar/barclamp-crowbar/pull/1254/files#diff-f125db84446056a04e9a58bb6052128cR45

or maybe you see a different scenario, where this would fail?

crowbar_framework/app/models/proposal.rb
bkutil: > Probably not something for now, but we should definitely have an item in the EPIC tasklist for getting rid of these stupid IDs. The `bc-` prefix is not needed.

Added to the EPIC card.

crowbar_framework/app/models/proposal.rb
vuntz: Okay, let's keep the same list for now, but maybe create a task for later to revisit this once the move to Proposal is over.

(and happy to hear that the template doesn't end up in the db :-))

crowbar_framework/app/models/proposal_object.rb
vuntz: Was just checking that it's hashes, so we're good :-)

crowbar_framework/db/migrate/20150312081459_create_proposals.rb
bkutil: > * the revisions and crowbar-applied bits that are in the deployment part should likely just be straight columns in the database anyway

I agree, if we want to keep the revisions, they should be in a column.

> That being said, I'd definitely agree with doing that as a later step; so a research card is fine with me.

Ok, thanks for the explanation. I created a task in the EPIC card for the
research.

crowbar_framework/app/models/proposal.rb
bkutil: I added a test to make sure that does not happen. The check is based on 'template' being a reserved name. We could also compare the `properties` and don't save if they're unchanged from the proposal template. Would that be better?

crowbar_framework/db/migrate/20150312081459_create_proposals.rb
tboerger: And on top of that deployment and attributes is already handled more or less differntly, even when it is displayed on a single page :)

tboerger:
I think it looks pretty good now. I have not tested it but codewise i'm fine with that. :+1:

MaximilianMeister:
@bkutil How can I test this? do the patches need to be applied before crowbar installation? or is it sufficient to pull the patches into a running admin node?

bkutil:
@MaximilianMeister you can do either of those two, I think. Before crowbar install would be better (as we'd know the patches don't break it). You just need to remove hunks that touch specs, as the files are not installed on the admin node.

MaximilianMeister:
@bkutil okay, maybe i'll test both. and if i pull it into a running system, do i need to do some `rake db:migrate` or sth similar?

bkutil:
@MaximilianMeister right, rake db:migrate is needed to create the proposals table.

MaximilianMeister:
don't know if that belongs to this code/PR but how will the proposals be migrated over to the database in terms of a cloud version update?
i was pulling the patches into a running admin node, and right now only new proposals are in the sqlite db.

bkutil:
> don't know if that belongs to this code/PR but how will the proposals be migrated over to the database in terms of a cloud version update?
> i was pulling the patches into a running admin node, and right now only new proposals are in the sqlite db.

Yep, that's the plan. I think the upgrade is a separate task in the
epic.

MaximilianMeister:
I tested the functionality before, and after installing crowbar.
* creating, changing, deleting, showing proposals from the crowbar CLI + UI
* looking in the database if all is fine

so if i did not miss any test:

:+1: QA wise

otherwise please tell me if there is anything else that needs to be tested

i only found an issue during deployment of nova, but afaics this not related to this code
http://pastebin.suse.de/13353/src

bkutil:
@MaximilianMeister no, I think that should be all from the QA side

crowbar_framework/app/models/proposal.rb
vuntz: Hrm, I think it can happen that we save a proposal that is unchanged compared to the template (or is something always changing?)

Anyway, current code is fine.

vuntz:
+1

crowbar_framework/app/models/proposal.rb
bkutil: Yes, we can save a proposal, whose 'properties' - attributes/deployment are unchanged from the template, but it will have a different name (i.e. not 'template').

PR #1255, opened 2015-03-20
[backport] Make NodeObject.group_order return a unique value

vuntz:
The node dashboard relies on the fact that group_order is different for
each node. This was not the case when two nodes are actually VMWare VMs
that are ultimately plugged in the same switch.

So instead of only relying on the information of where the node is
plugged on the switch, also append to that the alias of the node to
ensure group_order is unique.

(cherry picked from commit f40c7404030ea069f64bca4186b33a04cf5bb2c9)


tboerger:
:+1:

PR #1256, opened 2015-03-24
remove broken collapse menu on port 80 page

thutterer:
As discussed [here](https://github.com/SUSE-Cloud/barclamp-crowbar/pull/18) we don't need the collapse menu as we only have a single link in the right navbar menu and don't want to depend on more CDN content.


tboerger:
:+1:

bkutil:
+1

PR #1257, opened 2015-03-24
add and adjust links on port 80 page

thutterer:
As @nkrinner suggested in [Bugzilla](https://bugzilla.suse.com/show_bug.cgi?id=922699) the panel-heading text could be a link to port 3000 as well.
This needs minor changes in the page's JavaScript.

Besides that, I found two other small issues:
- The top-left logo is a link to "/" which does nothing but lead to the same error page the user is already on. So shouldn't we change that to a link to port 3000 too?
- The JavaScript adds ":3000" to the current location which is fine when you reached that site directly via port 80. But if you work with port-forwarding like in our virtual cloud setups, the link then becomes something like "hardware:1180:3000" which is not valid nor working. Shouldn't we improve the script to remove the trailing ":\d*" first and then add the ":3000" ? 
```javascript
var url = document.createElement('a');
url.href = window.location.host;
dashboard_link = url.hostname + ":3000";
```
That might be still the wrong link then but at least it is a valid one.

This PR was previously opened on [SUSE-Cloud](https://github.com/SUSE-Cloud/barclamp-crowbar/pull/19) but probably belongs here as @vuntz suggested.

### DIFF
--- a/crowbar_framework/index/index.html
+++ b/crowbar_framework/index/index.html
@@ -70,13 +70,13 @@
     <div class="navbar navbar-inverse navbar-fixed-top" role="navigation">
       <div class="container">
         <div class="navbar-header">
-          <a href="/" class="navbar-brand">
+          <a href="/" class="navbar-brand dashboard">
             <span class="title">Crowbar</span>
           </a>
         </div>
         <div class="navbar-right">
           <ul class="nav navbar-nav">
-            <li><a id="dashboard" href="http://127.0.0.1:3000/">Go to dashboard</a></li>
+            <li><a class="dashboard" href="http://127.0.0.1:3000/">Go to dashboard</a></li>
           </ul>
         </div>
       </div>
@@ -91,7 +91,7 @@
         </div>
         <div class="panel-body">
           <p>
-            The Crowbar Admin Server is only running on port 3000, so please follow the link above to the Crowbar dashboard. If
+            The Crowbar Admin Server is only running on <a class="dashboard" href="http://127.0.0.1:3000/">port 3000</a>. If
             you still have problems feel free to visit the <a href="http://github.com/crowbar/crowbar/wiki">Crowbar site</a> for
             more information.
           </p>
@@ -106,7 +106,10 @@
     </div>
 
     <script>
-      document.getElementById("dashboard").href="http://" + window.location.host + ":3000";
+      var links_to_dashboard = document.getElementsByClassName("dashboard");
+      for (i = 0; i < links_to_dashboard.length; i++) {
+        links_to_dashboard[i].href="http://" + window.location.host + ":3000";
+      }
     </script>
   </body>
 </html>


crowbar_framework/index/index.html
```diff
@@ -114,7 +116,10 @@
     </div>
 
     <script>
-      document.getElementById("dashboard").href="http://" + window.location.host + ":3000";
+      links_to_dashboard = document.getElementsByClassName("dashboard");
```
bkutil: The so called 'strict mode' in javascript raises exception on variables declared without `var` keyword. So while this works, it is better to write `var links_to_dashboard = `.

crowbar_framework/index/index.html
```diff
@@ -114,7 +116,10 @@
...
+      for (i = 0; i < links_to_dashboard.length; i++) {
+        links_to_dashboard[i].href="http://" + window.location.host + ":3000";
```
bkutil: Purely matter of style - the `i`th element can be assigned into a special variable. This is useful when the variable is accessed frequently inside the loop and also makes the code a tiny bit more readable:

```
var link = links_to_dashboard[i];
link.href = "...";
```

bkutil:
I also realized that what you can do within this patch is to not rely on `substring`, but use this technique to parse the url: http://stackoverflow.com/a/6168370 - that might help with the second point you mention.

> The top-left logo is a link to "/" which does nothing but lead to the same error page the user is already on. So shouldn't we change that to a link to port 3000 too?

Good idea.

> The JavaScript adds ":3000" to the current location which is fine when you reached that site directly ..(snipped)

If that'd fix the issue, go for it!

thutterer:
Thanks @bkutil! I have only very little expertise in JS so far and appreciate your tips a lot.
I changed my PR comment to reflect your idea about how to modify the URL. Reads way better now.
And I'll amend the missing `var` too.
But I'll skip the other tip about assigning the array elements to a separate variable as I only use it once in the loop. But in general a nice idea to improve readability.

tboerger:
There is already another pull request that redirects automatically to the dashboard 8)

vuntz:
I was playing with this, and I didn't see a huge difference -- until I noticed the title is a link. I think we should have the link in the text, not in the title.

(this needs to be rebased, btw)

tboerger:
And if we combine it with my suggestion on the other pull request it doesn't really matter because it gets automagically redirected :)

nkrinner:
@tboerger IMHO automated redirects are bad if they do more than fix a url typo or soemthing similar.

thutterer:
@tboerger ... and as long as there is no solution how to fix the problem with two port numbers in one link (s. my opening comment) it would try to redirect you to a url like "hostname:1181:3000" :/

tboerger:
@nkrinner Because of that i have set a delay of 5 seconds and changed the text related to that. https://github.com/crowbar/barclamp-crowbar/pull/1174#issuecomment-86049647

tboerger:
@thutterer And now the comment on the other pull requests redirects on ```hostname``` instead of ```host``` https://github.com/crowbar/barclamp-crowbar/pull/1174#issuecomment-86049647

thutterer:
@vuntz As discussed I've reworked the commit so that the link is now inside the help text and not in the heading.
@nkrinner As you opened the bug, is that ok for you?

nkrinner:
fine for me. +1

crowbar_framework/index/index.html
```diff
@@ -91,7 +91,7 @@
         </div>
         <div class="panel-body">
           <p>
-            The Crowbar Admin Server is only running on port 3000, so please follow the link above to the Crowbar dashboard. If
+          The Crowbar Admin Server is only running on <a class="dashboard" href="http://127.0.0.1:3000/">port 3000</a>. If
```
vuntz: Can you add back the leading spaces you removed? :-) Other than that, +1

thutterer: Done

tboerger:
Why did i get ignored so often? -.-

vuntz:
@tboerger this being merged doesn't mean that the automatic redirect will not make it. This patch was a quick and simple fix without any behavior change, so there was no reason to not merge it...

tboerger:
@vuntz I'm talking more about my suggestion related to ```host``` <=> ```hostname``` releated to appended port number...

vuntz:
> @vuntz I'm talking more about my suggestion related to ```host``` <=> ```hostname``` releated to appended port number...

Hrm, I guess this got lost in the review somehow. @thutterer was there any reason for not doing that?

thutterer:
This would avoid duplication of port numbers, yes. But only in virtual cloud setups and in such Crowbar does not run on port 3000 anyway and so the created link would still be wrong.
In a supported setup where the port80 page is on port80 and Crowbar is on 3000, it doesn't really make a difference ... although it may be a bit better JS style.
Maybe I've read @tboerger's message before he edited it to explain the difference ;)
But as long as we don't want to take care of redirection in a virtual cloud, I would leave it as it is for now.

PR #1260, opened 2015-03-31
[wip] Load proposal data from sqlite, instead via chef api

bkutil:
Needs testing, just checking if this is the correct approach.

cc: @vuntz @aspiers @tboerger

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -757,7 +757,7 @@ class ServiceObject
   # is computed (in apply_role) and chef client gets called on the nodes.
   # Hopefully, this will get moved into a background job.
   def proposal_commit(inst, in_queue = false, validate_after_save = true)
-    prop = ProposalObject.find_proposal(@bc_name, inst)
+    prop = Proposal.where(barclamp: @bc_name, name: inst).first
 
     if prop.nil?
       [404, "#{I18n.t('.cannot_find', :scope=>'model.service')}: #{@bc_name}.#{inst}"]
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/models/service_object.rb
```diff
@@ -769,7 +769,8 @@ def proposal_commit(inst, in_queue = false, validate_after_save = true)
         # Put mark on the wall
         prop["deployment"][@bc_name]["crowbar-committing"] = true
         save_proposal!(prop, :validate_after_save => validate_after_save)
-        response = active_update prop.raw_data, inst, in_queue
+        sqlite_prop = Proposal.where(:barclamp => @bc_name, :name => inst).first || raise("Failed to find proposal #{@bc_name}-#{inst}.")
```
vuntz: Localization of error message ;-)

vuntz:
You might want to clarify the commit message (to mention that this is about committing).

Would it be possible to remove the call at https://github.com/crowbar/barclamp-crowbar/pull/1260/files#diff-0be8ea1f05c9441a59bb2b7ac65b1c62R760 ? Because in the end, I think this is really what we should target here.

bkutil:
@vuntz we change the proposal afterwards (save_proposal!) and the changes would be made to `Proposal` instead of `ProposalObject` record. But if #1261 is merged, the models are synced both ways, so it should work then.

aspiers:
Agree with @vuntz, explanations belong in the commit message not as comments in the PR.

bkutil:
Updated the commit message. Merging still depends on #1261.

tboerger:
Failing tests :(

tboerger:
Awesome when the tests work! :)

bkutil:
@tboerger they should pass now :)

tboerger:
:+1:

vuntz:
+1

PR #1261, opened 2015-03-31
Bidirectional sync for models

bkutil:
This enables almost seamless replacement of ProposalObject and Proposal,
as the changes will be mirrored into the other model transparently.

The changes can be dropped after we're done migrating barclamps.

### DIFF
--- a/crowbar_framework/app/models/proposal.rb
+++ b/crowbar_framework/app/models/proposal.rb
@@ -22,8 +22,8 @@ class Proposal < ActiveRecord::Base
 
   # FIXME: these are safe to remove once all barclamps are converted to use
   # Proposal instead of ProposalObject
-  after_destroy :drop_corresponding_proposal_object
-  after_save    :update_corresponding_proposal_object
+  before_destroy :drop_corresponding_proposal_object
+  before_save    :update_corresponding_proposal_object
 
   # XXX: a 'registered' barclamp could have a has_many :proposals and have a factory
   # method for creating them. Then the check for barclamp arg would not be needed,
--- a/crowbar_framework/app/models/proposal.rb
+++ b/crowbar_framework/app/models/proposal.rb
@@ -101,7 +101,7 @@ class Proposal < ActiveRecord::Base
         proposal_object.save(sync: false)
       end
     else
-      bag = Chef::DataBagItem.json_create({"raw_data" => self.properties, "data_bag" => "crowbar"})
+      bag = Chef::DataBagItem.json_create({"raw_data" => self.raw_data, "data_bag" => "crowbar"})
       proposal_object = ProposalObject.new(bag)
       proposal_object.save(sync: false)
     end
--- a/crowbar_framework/app/models/proposal.rb
+++ b/crowbar_framework/app/models/proposal.rb
@@ -20,6 +20,11 @@ class Proposal < ActiveRecord::Base
   after_initialize :load_properties_template, :set_default_name
   before_save      :update_proposal_id
 
+  # FIXME: these are safe to remove once all barclamps are converted to use
+  # Proposal instead of ProposalObject
+  after_destroy :drop_corresponding_proposal_object
+  after_save    :update_corresponding_proposal_object
+
   # XXX: a 'registered' barclamp could have a has_many :proposals and have a factory
   # method for creating them. Then the check for barclamp arg would not be needed,
   # as we'd always know the barclamp exists.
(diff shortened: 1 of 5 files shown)


crowbar_framework/app/models/proposal_object.rb
```diff
@@ -126,14 +126,14 @@ def save(options = {})
     increment_crowbar_revision!
     Rails.logger.debug("Saving data bag item: #{@item["id"]} - #{crowbar_revision}")
     @item.save
-    save_proposal_in_sqlite
+    save_proposal_in_sqlite unless options.fetch(:skip_sync, false)
```
MaximilianMeister: is there any case where we would have `:skip_sync, false`? it does not hurt but I don't see the necessity here if it's true in all cases.
and maybe my notion of the sync is wrong but shouldn't it just be called `sync`? with `skip_sync, true` it should not sync

bkutil:
@MaximilianMeister I tried to make the option a bit clearer. Is this better?

MaximilianMeister:
@bkutil yes thanks. the easier to read, the better :) you can squash the commit with the first one if you like, i guess it was just for showing the diff for me :)

bkutil:
@MaximilianMeister squashed

bkutil:
Tested this by

1. patching barclamp controller to load `Proposal` instead of `ProposalObject`
2. modifying nfs barclamp's proposal (assigning a node)
3. checking in rails console that the changes propagated to the other model.

aspiers:
Sounds very cool and looks good at a quick glance.

crowbar_framework/app/models/proposal.rb
```diff
@@ -87,6 +92,26 @@ def raw_data=(value)
...
+  # XXX: we need to be careful to not create an endless loop
+  def update_corresponding_proposal_object
+    proposal_object = ProposalObject.find_proposal_by_id(self.key)
+    if proposal_object
+      if proposal_object.raw_data != self.raw_data
+        proposal_object.raw_data = self.raw_data
+        proposal_object.save(sync: false)
+      end
+    else
+      bag = Chef::DataBagItem.json_create({"raw_data" => self.properties, "data_bag" => "crowbar"})
```
vuntz: Why not `self.raw_data`?

crowbar_framework/app/models/proposal.rb
```diff
@@ -87,6 +92,26 @@ def raw_data=(value)
 
   private
 
+  # XXX: we need to be careful to not create an endless loop
+  def update_corresponding_proposal_object
+    proposal_object = ProposalObject.find_proposal_by_id(self.key)
+    if proposal_object
+      if proposal_object.raw_data != self.raw_data
+        proposal_object.raw_data = self.raw_data
+        proposal_object.save(sync: false)
```
vuntz: Hrm, don't we end up with some discrepancy in the revision between Proposal and ProposalObject (since it'll be bumped in this save again)?

crowbar_framework/app/models/proposal.rb
bkutil: They are equivalent. I'll change this to be consistent in the method.

crowbar_framework/app/models/proposal.rb
bkutil: Good catch. I'll fix this and give it a test.

bkutil:
@vuntz the crowbar-revision will now be synced properly. tested manually in both directions (Proposal object to Proposal sync and back)

tboerger:
:+1:

crowbar_framework/app/models/proposal.rb
```diff
@@ -87,6 +92,27 @@ def raw_data=(value)
 
   private
 
+  # XXX: we need to be careful to not create an endless loop
+  def update_corresponding_proposal_object
+    proposal_object = ProposalObject.find_proposal_by_id(self.key)
+    if proposal_object
+      if proposal_object.raw_data != self.raw_data
+        increment_crowbar_revision! if proposal_object.crowbar_revision >= self.crowbar_revision
```
vuntz: I'm a bit confused here:
- first, shouldn't we always increment the revision before saving (unless we do the save for a sync, of course)?
- I'm not sure what the `if` condition is trying to protect against?

nkrinner:
Looks good to me. +1

crowbar_framework/app/models/proposal.rb
bkutil: > first, shouldn't we always increment the revision before saving (unless we do the save for a sync, of course)?
I'll go through the code again. I'm sure there is a reason for this, but I no longer remember ;)

> I'm not sure what the if condition is trying to protect against?

Which one?

bkutil: Ok, you were right. The `if` for revision compare is not needed. I dropped it. Still not sure about the 'always increment', though. Where should be the incrementing method called?

vuntz:
+1

PR #1263, opened 2015-04-23
Use admin name attribute when computing host for admin URL (bsc#927469)

vuntz:
This is the exact same thing we're doing for the public IP address/DNS
name.

The reason it's needed is for keystone admin API which are being done
through the admin endpoint. As this is a rather minimal use case, we do
not expose the setting in the UI for the nodes -- which means the node
would have to be edited directly.

https://bugzilla.suse.com/show_bug.cgi?id=927469

Depends on https://github.com/crowbar/barclamp-pacemaker/pull/190


mjura:
I haven't tested this but looks cool to me +1

tboerger:
LGTM

PR #1264, opened 2015-04-27
[wip] Replace calls to ProposalObject with Proposal

bkutil:
This patchset replaces most of the calls to ProposalObject with calls to Proposal in the crowbar barclamp.

What has not been replaced and is still backed by databags:

  * The deployment queue
  * Network configuration

### DIFF
--- a/crowbar_framework/app/controllers/barclamp_controller.rb
+++ b/crowbar_framework/app/controllers/barclamp_controller.rb
@@ -175,7 +175,7 @@ class BarclampController < ApplicationController
     return render :text => @proposals, :status => ret[0] if ret[0] != 200
     respond_to do |format|
       format.html {
-        @proposals.map! { |p| Proposal.where(barclamp: @bc_name, name: p).first }
+        @proposals = @proposals.map { |p| Proposal.where(barclamp: @bc_name, name: p).first }
         render :template => 'barclamp/proposals'
       }
       format.xml  { render :xml => @proposals }
--- a/crowbar_framework/app/controllers/barclamp_controller.rb
+++ b/crowbar_framework/app/controllers/barclamp_controller.rb
@@ -175,7 +175,7 @@ class BarclampController < ApplicationController
     return render :text => @proposals, :status => ret[0] if ret[0] != 200
     respond_to do |format|
       format.html {
-        @proposals.map! { |p| ProposalObject.find_proposal(@bc_name, p) }
+        @proposals.map! { |p| Proposal.where(barclamp: @bc_name, name: p).first }
         render :template => 'barclamp/proposals'
       }
       format.xml  { render :xml => @proposals }
@@ -238,7 +238,6 @@ class BarclampController < ApplicationController
   def get_proposals_from_barclamps(barclamps)
     modules = {}
     active = RoleObject.active
-    proposals = ProposalObject.all
     barclamps.each do |name, details|
       modules[name] = { :description=>details['description'] || t('not_set'), :order=> details['order'], :proposals=>{}, :expand=>false, :members=>(details['members'].nil? ? 0 : details['members'].length) }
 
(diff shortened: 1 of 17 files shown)


bkutil:
Tested by patching an admin node and applying / saving / modifying the NFS barclamp.

crowbar_framework/app/models/service_object.rb
```diff
@@ -604,17 +604,17 @@ def element_info(role = nil)
...
 
   def proposals_raw
-    ProposalObject.find_proposals(@bc_name)
+    Proposal.where(barclamp: @bc_name)
   end
 
   def proposals
     props = proposals_raw
-    props.map! { |p| p["id"].gsub("bc-#{@bc_name}-", "") } unless props.empty?
+    props = props.map { |p| p["id"].gsub("bc-#{@bc_name}-", "") } unless props.empty?
```
vuntz: I know it's not your fault, but maybe we can drop the useless `unless` here?

vuntz:
Need to look at it tomorrow, but after a quick glance, looks good.

bkutil:
@vuntz I removed the extra `unless x.empty?`

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -369,9 +367,8 @@ def proposal_update
       return render :text => ret[1], :status => ret[0] if ret[0] != 200
       return render :json => ret[1]
     else # This is UI.
-      params[:id] = "bc-#{params[:barclamp]}-#{params[:id] || params[:name]}"
       if params[:submit] == t('barclamp.proposal_show.save_proposal')
-        @proposal = ProposalObject.find_proposal_by_id(params[:id])
+        @proposal = Proposal.where(barclamp: params[:barclamp], name: params[:id] || params[:name]).first
```
vuntz: I'm a bit unsure here: it seems `params[:name]` would be the full name (`bc-barclamp-foo`), in which case I don't think it's correct to use it that way.

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -382,7 +379,7 @@ def proposal_update
           flash_and_log_exception(e)
         end
       elsif params[:submit] == t('barclamp.proposal_show.commit_proposal')
-        @proposal = ProposalObject.find_proposal_by_id(params[:id])
+        @proposal = Proposal.where(barclamp: params[:barclamp], name: params[:id] || params[:name]).first
```
vuntz: (same comment)

vuntz:
@bkutil did you also test creating a new proposal with the patch?

crowbar_framework/app/controllers/barclamp_controller.rb
bkutil: Looking at how the `params[:id]` is constructed on line [372](https://github.com/crowbar/barclamp-crowbar/pull/1264/files#diff-6e0801372d9a358289edc702ef124633L372) - it seems that this should continue working as expected?

bkutil:
@vuntz yes. I created a new NFS proposal, then added nodes, saved, then deployed. I'll try to test the whole admin node installation process today together w/ Bernhard.

crowbar_framework/app/controllers/barclamp_controller.rb
vuntz: Doh, sorry, I had misread the code :-)

vuntz:
+1

crowbar_framework/app/controllers/barclamp_controller.rb
bkutil: No worries, it is indeed weird...

MaximilianMeister:
+1

cool, didn't know about `with_indifferent_access` yet.

mjura:
+1

PR #1265, opened 2015-05-04
Allow passing in logger as an option to RemoteNode methods

bkutil:
RemoteNode uses puts as a default way of logging (since it assumes it
  will be called from a forked process, where all file descriptors except
  for stdout/in/err are closed).
    
  However, it now needs to be used in another context, namely to check if
  chef clients are running prior to applying of the role.
    
  This patch adds an options hash to relevant methods and allows passing a
  logger option. If not specified, it falls back to using puts.

cc @vuntz, @MaximilianMeister #1251

### DIFF
--- a/crowbar_framework/lib/remote_node.rb
+++ b/crowbar_framework/lib/remote_node.rb
@@ -24,7 +24,7 @@ module RemoteNode
   # @param host [string] hostname or IP
   # @param timeout [integer] timeout in seconds
 
-  def self.ready?(host, timeout = 60, sleep_time = 10)
+  def self.ready?(host, timeout = 60, sleep_time = 10, options = {})
     timeout_at = Time.now + timeout
     ip         = self.resolve_host(host)
     msg = "Waiting for host, next attempt in #{sleep_time} seconds until #{timeout_at}"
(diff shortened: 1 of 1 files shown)


crowbar_framework/lib/remote_node.rb
```diff
@@ -64,11 +64,11 @@ def self.chef_clients_running?(host_or_ip)
     end
   end
 
-  def self.reboot_done?(host_or_ip, reboot_requesttime)
+  def self.reboot_done?(host_or_ip, reboot_requesttime, options)
```
MaximilianMeister: why don't we have a default value for `options` here?

crowbar_framework/lib/remote_node.rb
```diff
@@ -120,13 +122,16 @@ def self.ssh_cmd_base(host_or_ip)
   end
 
   # get boot time of the given host or 0 if command can not be executed
-  def self.ssh_cmd_get_boottime(host_or_ip)
+  def self.ssh_cmd_get_boottime(host_or_ip, options)
```
MaximilianMeister: same here

MaximilianMeister:
BTW Thanks for the quick rescue for https://github.com/crowbar/barclamp-crowbar/pull/1251

crowbar_framework/lib/remote_node.rb
MaximilianMeister: if not needed, we will have to adapt the [spec](https://github.com/crowbar/barclamp-crowbar/blob/master/crowbar_framework/spec/lib/remote_node_spec.rb) i guess

bkutil:
@MaximilianMeister thanks, that is a good point. I originally thought that since those two methods (`reboot_done?` and `wait_until`) are private, they should require the options argument (as they won't be called from anywhere else). But since for ruby, private does not mean 'cannot be called from outside', I think you're right and it is better to be on the safe side.

I updated the code accordingly.

tboerger:
:+1:

MaximilianMeister:
for me that looks good as a solution, @vuntz what do you think in relation to https://github.com/crowbar/barclamp-crowbar/pull/1251?

vuntz:
Didn't test it, but assuming it works, +1

MaximilianMeister:
I have just tested it in combination with https://github.com/crowbar/barclamp-crowbar/pull/1251, and it works just fine
:+1:

PR #1267, opened 2015-05-06
Ignore test coverage output and ruby switching specific files

bkutil:


aspiers:
+1 for ignoring `coverage/`.  `.ruby-version` I'm not so convinced about, because even if we ignore `Gemfile.lock`, there could easily be dependencies in `Gemfile` which rely on a particular Ruby version, and I guess the [bundler we use probably still fails to cope with this](https://github.com/bundler/bundler/issues/2965).

So I'd recommend to split these into different commits, and maybe even different PRs.

bkutil:
> .ruby-version I'm not so convinced about, because even if we ignore Gemfile.lock, there could easily be dependencies in Gemfile which rely on a particular Ruby version,

Can you please elaborate? I'm not sure how git ignoring `.ruby-version` affects gems which require a certain Ruby.

vuntz:
FWIW, +1 on whatever decision you take in this discussion :-)

aspiers:
>  I'm not sure how git ignoring .ruby-version affects gems which require a certain Ruby.

Because if `.ruby-version` is not in `git` then people could waste time trying to use Ruby versions which are already known not to work, e.g. `debugger` vs. `ruby-debug19`, or `jazz_hands` vs. `jazz_fingers`.

Having said that, do we want to support multiple ruby versions for a given release?  If so then I guess we would have to avoid checking in `.ruby-version`, since IIRC it only allows specifying one version.

bkutil:
Hm, but `.ruby-version` is not in git yet. So you're suggesting that we should add it?

aspiers:
Well, either we should add it, *or* add it to `.gitignore`.  Which depends on the answer to my question above.

bkutil:
> Having said that, do we want to support multiple ruby versions for a given release?

I don't think that is supported. Maybe @vuntz has more info on that?

vuntz:
Let's put it that way: I don't think we want to discourage supporting multiple versions, but we don't plan to proactively do that.

bkutil:
OK, let's leave the gitignore as it is now. Closing.

aspiers:
@bkutil I don't see why closing is an appropriate route to take:

1. We still need to ignore `coverage/`
2. Like I said, either we should add `.ruby-version` to the repo, or add it to `.gitignore`

@vuntz's last comment implies that we *do* want to allow other people to try different Ruby versions (even if at SUSE we only stick to a single version), therefore it makes sense to add it to `.gitignore` and so I would have given this PR a +1 if it wasn't closed.

tboerger:
Please reopen this PR and let's merge that in. IMHO it makes sense like it is.

PR #1269, opened 2015-05-06
Replace remaining ProposalObject calls with Chef::DataBag

bkutil:
This patch replaces the remaining uses of ProposalObject with either Proposal calls, or, where none of the Proposal API methods are required (as e.g. for the deployment queue), uses Chef::DataBag class directly.

This will allow for dropping the ProposalObject related code now, then move away from data bags towards a separate table/database later.

There are similar pull requests for other barclamps, they'll be referenced below.

I noticed a slight problem with this and referencing PRs - ProposalObject rescues any error raised by Chef::DataBag.load - this'll have to be fixed...

### DIFF
--- a/crowbar_framework/app/models/proposal_object.rb
+++ b/crowbar_framework/app/models/proposal_object.rb
@@ -140,6 +140,8 @@ class ProposalObject < ChefObject
   private
 
   def delete_proposal_from_sqlite
+    return if self.id =~ /_network/
+
     attrs = { barclamp: self.barclamp, name: self.name }
 
     prop = Proposal.where(attrs).first
@@ -147,6 +149,8 @@ class ProposalObject < ChefObject
   end
 
   def save_proposal_in_sqlite
+    return if self.id =~ /_network/
+
     attrs = { barclamp: self.barclamp, name: self.name }
 
     prop = Proposal.where(attrs).first_or_initialize(attrs)
--- a/crowbar_framework/app/models/proposal_object.rb
+++ b/crowbar_framework/app/models/proposal_object.rb
@@ -140,6 +140,8 @@ class ProposalObject < ChefObject
   private
 
   def delete_proposal_from_sqlite
+    # XXX: Leave network proposals alone - its a network configuration
+    # this is transitional code, as soon as network uses databags, this can go away.
     return if self.id =~ /_network/
 
     attrs = { barclamp: self.barclamp, name: self.name }
@@ -149,6 +151,8 @@ class ProposalObject < ChefObject
   end
 
   def save_proposal_in_sqlite
+    # XXX: Leave network proposals alone - its a network configuration
+    # this is transitional code, as soon as network uses databags, this can go away.
     return if self.id =~ /_network/
 
     attrs = { barclamp: self.barclamp, name: self.name }
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1036,7 +1036,7 @@ class ServiceObject
   end
 
   def _proposal_update(bc_name, inst, proposal, validate_after_save = true)
-    prop = Proposal.new(barclamp: bc_name, name: inst)
+    prop = Proposal.where(barclamp: bc_name, name: inst).first_or_initialize(barclamp: bc_name, name: inst)
 
     begin
       prop.properties = proposal
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -776,9 +776,10 @@ class ServiceObject
         response = [500, e.message]
       ensure
         # Make sure we unmark the wall
-        prop = ProposalObject.find_proposal(@bc_name, inst)
+        prop = Proposal.where(barclamp: @bc_name, name: inst).reload
         prop["deployment"][@bc_name]["crowbar-committing"] = false
-        prop.save(:applied => (response.first == 200))
+        prop.latest_applied = (response.first == 200)
+        prop.save
       end
       response
     end
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -287,7 +287,7 @@ class ServiceObject
     begin
       f = acquire_lock "queue"
 
-      db = Chef::DataBag.load "crowbar/queue"
+      db = Chef::DataBag.load("crowbar/queue") rescue nil
       if db.nil?
         db = Chef::DataBagItem.new
         db.data_bag "crowbar"
@@ -385,7 +385,7 @@ class ServiceObject
     begin
       f = acquire_lock "queue"
 
-      db = Chef::DataBag.load "crowbar/queue"
+      db = Chef::DataBag.load("crowbar/queue") rescue nil
       @logger.debug("dequeue proposal: exit #{inst} #{bc}: no entry") if db.nil?
       return [200, {}] if db.nil?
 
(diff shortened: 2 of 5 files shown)


bkutil:
I noticed a slight problem with this and referencing PRs - ProposalObject rescues any error raised by Chef::DataBag.load - this'll have to be fixed...

crowbar_framework/app/models/service_object.rb
```diff
@@ -722,14 +725,14 @@ def proposal_create(params)
     # When we create a proposal, it might be "invalid", as some roles might be missing
     # This is OK, as the next step for the user is to add nodes to the roles
     # But we need to skip the after_save validations in the _proposal_update
-    _proposal_update(proposal, false)
+    _proposal_update(@bc_name, params["id"] || params[:name], proposal, false)
```
vuntz: Hrm, it seems the beginning of the method assumes `params["id"]` is always non-nil. Not that it matters much (and I think your bit is actually more correct based on what everything else is doing)

Would it make sense to use the `base_id` variable instead? (and maybe change how it's defined)

crowbar_framework/app/models/service_object.rb
```diff
@@ -776,9 +779,10 @@ def proposal_commit(inst, in_queue = false, validate_after_save = true)
         response = [500, e.message]
       ensure
         # Make sure we unmark the wall
-        prop = ProposalObject.find_proposal(@bc_name, inst)
+        prop = Proposal.where(barclamp: @bc_name, name: inst).first.reload
```
vuntz: Why `reload`?

crowbar_framework/app/models/service_object.rb
```diff
@@ -722,14 +725,14 @@ def proposal_create(params)
...
-    _proposal_update(proposal, false)
+    _proposal_update(@bc_name, params["id"] || params[:name], proposal, false)
   end
 
   def proposal_edit(params, validate_after_save = true)
     params["id"] = "bc-#{@bc_name}-#{params["id"] || params[:name]}"
     proposal = {}.merge(params)
     clean_proposal(proposal)
-    _proposal_update(proposal, validate_after_save)
+    _proposal_update(@bc_name, params["id"] || params[:name], proposal, validate_after_save)
```
vuntz: Maybe use a variable for `params["id"] || params[:name]` since it's used twice?

crowbar_framework/app/models/service_object.rb
bkutil: > Would it make sense to use the `base_id` variable instead? (and maybe change how it's defined)

Good catch. This is actually a bug. The base_id *needs* to be used.

crowbar_framework/app/models/service_object.rb
bkutil: > Why `reload`?

The repeated find on ProposalObject fetched all the changes made since
the first call at the top of the method.

So reload does the same - otherwise, rails would see that it already
fetched the model and simply re-save it, IIUC.

crowbar_framework/app/models/service_object.rb
bkutil: > Maybe use a variable for `params["id"] || params[:name]` since it's used twice?

Again, this is a bug - only the original `params["id"]` is needed.

tboerger:
Nice :+1:

crowbar_framework/app/models/service_object.rb
```diff
@@ -722,14 +725,15 @@ def proposal_create(params)
     # When we create a proposal, it might be "invalid", as some roles might be missing
     # This is OK, as the next step for the user is to add nodes to the roles
     # But we need to skip the after_save validations in the _proposal_update
-    _proposal_update(proposal, false)
+    _proposal_update(@bc_name, base_id, proposal, false)
   end
 
   def proposal_edit(params, validate_after_save = true)
-    params["id"] = "bc-#{@bc_name}-#{params["id"] || params[:name]}"
+    base_id = params["id"] || params[:name]
```
vuntz: So now I have to wonder: why do we not accept `params[:name]` in proposal_create and accept it here?

(this is not your change, but would be good to understand so we can fix it if needed)

crowbar_framework/app/models/service_object.rb
vuntz: Can't we simply do `prop.reload`?

crowbar_framework/app/models/service_object.rb
tboerger: If i remember correctly some legacy code uses ```name``` and newer routes/actions uses ```id```

crowbar_framework/app/models/service_object.rb
bkutil: Good point. Done.

crowbar_framework/app/models/service_object.rb
vuntz: @tboerger I understand that. But that doesn't explain the inconsistency between proposal_create and proposal_edit :-)

bkutil:
Unfortunately, there are still some problems (which I'm looking into ATM)

* I'm seeing the following exception in the logs:

```
E, [2015-05-11T17:48:29.616174 #16102] ERROR -- : SQLite3::BusyException: database is locked: UPDATE "proposals" SET "properties" = ? WHERE "proposals"."id" = 3
F, [2015-05-11T17:48:29.617260 #16102] FATAL -- : json/transition for ipmi:default failed: SQLite3::BusyException: database is locked: UPDATE "proposals" SET "properties" = ? WHERE "proposals"."id" = 3
```

* Testing fails with:

```
+ ssh crowbar.c9.cloud.suse.de 'dig multi-dns.'\''c9.cloud.suse.de'\'' | grep -q 10.11.12.13'
+ complain 13 'Multi DNS server test failed!'
```

* The proposals do not show as applied even on a fresh admin node install.

crowbar_framework/app/models/service_object.rb
tboerger: > @tboerger I understand that. But that doesn't explain the inconsistency between proposal_create and proposal_edit :-)

I can not say where this comes from, but this is something that can be resolved :)

bkutil:
> I'm seeing the following exception in the logs:

Using [immediate transaction mode](https://www.sqlite.org/lang_transaction.html) of sqlite, i.e. grabbing a reserved lock on begin transaction immediately, preventing other processes/threads from doing the same, seems to mitigate the issue. I was getting the exception consistently before patching and it did not appear on two tests afterwards. This is as far from a proof as it can be, but still worth trying. It requires a single line monkey patch to the SQlite adapter in rails (an API which did not change since rails 2.3).

bkutil:
> Testing fails with

Seems to fail regardless of whether code from this branch or staging is used. I'll need to investigate, or ask @bmwiedemann for advice.

bkutil:
Test setup now passes.

tboerger:
Still +1 :)

crowbar_framework/app/models/proposal_object.rb
```diff
@@ -140,6 +140,8 @@ def destroy(options = {})
   private
 
   def delete_proposal_from_sqlite
+    return if self.id =~ /_network/
```
vuntz: Maybe add a `XXX` comment here like you did in proposal.rb?

crowbar_framework/app/models/proposal_object.rb
```diff
@@ -147,6 +149,8 @@ def delete_proposal_from_sqlite
   end
 
   def save_proposal_in_sqlite
+    return if self.id =~ /_network/
```
vuntz: (here too)

rsalevsky:
:+1:

crowbar_framework/app/models/service_object.rb
bkutil: @vuntz modification of `params[:id]` so that it falls back to / checks the value of `params[:name]` is done in [barclamp controller](https://github.com/crowbar/barclamp-crowbar/blob/master/crowbar_framework/app/controllers/barclamp_controller.rb#L331) instead. So a proper fix would be to have both assignments (proposal_update and proposal_create) either in the model, or in the controller.

PR #1271, opened 2015-05-11
Validate on migrate

toabctl:


toabctl:
This replaces https://github.com/crowbar/barclamp-crowbar/pull/1123 and most of the comments are addressed.

toabctl:
@tboerger you added the comments to the original PR. Is this fine now? @vuntz your chance to add a +1 to your own code :)

bkutil:
Improving migration robustness looks like a good idea.

One thing that I noticed is, that the 'data bags path' keeps popping up on many places. It would be nice to create a trello card to make this path a part of some configuration file or an initializer and make the code use it consistently.

vuntz:
Thanks so much for picking that up!

+1

crowbar_framework/lib/schema_migration.rb
```diff
@@ -143,7 +152,11 @@ def self.migrate_object(bc_name, template, all_scripts, attributes, deployment)
     scripts.each do |script|
       # we only pass attributes and deployment to not encourage direct access
       # to the proposal
-      attributes, deployment = run_script(script, is_upgrade, template['attributes'][bc_name], template['deployment'][bc_name], attributes, deployment)
+      begin
+        attributes, deployment = run_script(script, is_upgrade, template['attributes'][bc_name], template['deployment'][bc_name], attributes, deployment)
+      rescue StandardError => e
+        raise "error while executing migration script #{script}:\n#{e.message}"
```
tboerger: ```Error```

tboerger:
Minimal nitpic, i like it! :+1:

crowbar_framework/lib/schema_migration.rb
vuntz: The reason it's lowercase is because of how the string is displayed later on.

tboerger: But all other ```raise``` messages are written with uppercase first letter?

vuntz: That's really because the other messages are actually displayed in this string (`e.message`), iirc. I don't remember the details for sure; the best way to figure it out is to test the patch :-)

rsalevsky:
+1

PR #1272, opened 2015-05-11
[tex] Use admin name attribute when computing host for admin URL (bsc#927469)

vuntz:
**Backport of https://github.com/crowbar/barclamp-crowbar/pull/1263 to tex**

This is the exact same thing we're doing for the public IP address/DNS
name.

The reason it's needed is for keystone admin API which are being done
through the admin endpoint. As this is a rather minimal use case, we do
not expose the setting in the UI for the nodes -- which means the node
would have to be edited directly.

https://bugzilla.suse.com/show_bug.cgi?id=927469
(cherry picked from commit 89628b8672a14200c52c7f274f3c8b318258d4c1)


jsuchome:
Looks good +1

vuntz:
Backport of something in master, so one +1 is good enough to merge.

PR #1273, opened 2015-05-13
Use immediate transaction mode in SQLite

bkutil:
With deferred transaction mode, used by ActiveRecord by default, we were
getting SQLite3::BusyException when two processess attempted to do an update to
IPMI proposal at the same time. Using immediate transaction mode
(https://www.sqlite.org/lang_transaction.html) prevents two threads/processess
to begin an (immediate) transaction simultaneously and seems to alleviate the
issue.

refs https://github.com/crowbar/barclamp-crowbar/pull/1269

### DIFF
--- a/crowbar_framework/config/initializers/sqlite_transaction_monkey_patch.rb
+++ b/crowbar_framework/config/initializers/sqlite_transaction_monkey_patch.rb
@@ -1,0 +1,32 @@
+#
+# Copyright 2015, SUSE LINUX Products GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


vuntz:
I'll trust you :-) +1

bkutil:
We'll see. If the issue appears again, we can switch to a different DB. Trying this is just easier for now.

aspiers:
+1 although I'm surprised there isn't some driver option for doing this.

tboerger:
http://stackoverflow.com/questions/78801/sqlite3busyexception

```
production:
  adapter: sqlite3
  database: db/mysite_prod.sqlite3
  timeout: 10000
```

> The timeout value is specified in miliseconds. Increasing it to 10 or 15 seconds should decrease the number of BusyExceptions you see in your log.
>
> This is just a temporary solution, though. If your site needs true concurrency then you will have to migrate to another db engine.

tboerger:
So if https://github.com/crowbar/barclamp-crowbar/pull/1273#issuecomment-101644698 works this is a :-1: from my side.

bkutil:
I stumbled upon that thread as well, tested that change, but unfortunately, it did not help. I actually wanted to add it as a second patch, but then decided against it, as this change alone seemed sufficient.

bkutil:
@aspiers yep, maybe because it is a part of the transaction sql syntax?

tboerger:
Than I'm fine with that :)

PR #1275, opened 2015-05-25
Replace chef server with sqlite as a backend for proposals

bkutil:
Introduce a new Proposal ActiveRecord model, backed by sqlite database, then transition the code to use it instead of ProposalObject. Other barclamps have similarly named branches with the same purpose.

All the changes have been already reviewed before being merged into chef_db_drop feature branch.


zehicle:
just for the record... this is a similar path that was already walked with OpenCrowbar (for similar reasons).  We found that sqlite did not have the threading capabilities nor the JSON support needed.

bkutil:
@zehicle - interesting. Could you please elaborate on both points? I.e. lack of threading support and what JSON functionality was missing? (I hit an issue with concurrent transactions and tried to mitigate it in bb63006).

tboerger:
Awesome!
+1

tboerger:
> just for the record... this is a similar path that was already walked with OpenCrowbar (for similar reasons). We found that sqlite did not have the threading capabilities nor the JSON support needed.

Problems with transactions are already a known problem as mentioned by @bkutil, but that can be solved. We are not limited to sqlite, we can switch to which ever database we want :).

I can not see any problem with JSON as the workload gets anyway handled on the application and not on the database.

vuntz:
Assuming I gave a +1 to all changes in there already, you can consider my +1 is already here.

Do we need to merge the similarly-named branch in all other barclamps too at the same time?

bkutil:
> Do we need to merge the similarly-named branch in all other barclamps too at the same time?

Yes, or this one first and the other ones later.

bkutil:
Actually, as #1276 got merged, all PRs need to be merged at once.

tboerger:
Awesome!!!

PR #1276, opened 2015-05-26
Remove ProposalObject model and associated sync methods

bkutil:


aspiers:
Travis failure aside, :-1: since there is no commit message body.  That's always needed, but even more so when making fundamental architectural changes.  I realise that this PR is against a feature branch rather than `master`, but I don't think that justifies breaking the rule.

tboerger:
Beside the failing tests I'm fine with that. IMHO it does not require further explanation.

aspiers:
This looks much better now, thanks!  However I think you still need to adjust `bin/crowbar_reset`, which will break due to using `knife` rather than the `crowbar_*` CLI / API (yes that's dumb and it's my fault - sorry).

bkutil:
@aspiers thats a good point. Do you know if there are any other scripts/places in code, that could be using knife instead of the API?

aspiers:
I can't see any off-hand, but I guess we'll notice if something breaks ...  In contrast, when we move the nodes model to the database, we'll have to do more (try `grep -r knife /opt/dell/bin`).

bkutil:
@aspiers could you please help me a bit? which API operations correspond to the knife + options passed to it?

aspiers:
Hmm good point, actually there aren't any APIs for fixing a broken state machine ;-)  So I guess this needs to be ported to manually frob the SQL database instead.

bkutil:
Ok, same question :) I'm not sure, what do the knife commands do? What
would be an equivalent SQL/ActiveRecord query?

tboerger:
@bkutil You have to edit the attributes JSON, for proposal reset you need to set the ```crowbar-committing``` attribute back to ```false```, and for nodes reset you need to set the role ```state``` attribute back to ```ready```, nothing more happens for that.

tboerger:
I think it should be fairly easy to integrate these methods into the API.

aspiers:
They shouldn't be in the API, they are workarounds for internal bugs.

bkutil:
@aspiers which bugs?

tboerger:
Than do some rake tasks? Or just port the existing bin script to interact with the models.

aspiers:
@bkutil https://bugzilla.suse.com/show_bug.cgi?id=840255

> Or just port the existing bin script to interact with the models.

Yes that was my suggestion too.  It could either directly tweak the SQL db, or do it via the Rails models, which would be a lot nicer.  I don't see a big advantage in providing that via Rake tasks over a crowbar CLI subcommand, though.

vuntz:
+1 for the current code, but based on the comments, there's still some bits left to complete.

bkutil:
@aspiers @tboerger Do you think adding something like this to the script would be sufficient?

```bash
cmd="p = Proposal.where(barclamp: '"$bc"', name: '"$proposal"').first || raise; p['deployment']['"$bc"']['crowbar-committing'] = false; p.save;"

pushd "$RAILS_ROOT"
RAILS_ENV=production rails r "$cmd"
popd
```

aspiers:
That sounds reasonable to me.  Of course it would be a lot nicer to ditch the bash and do this in pure Ruby, but for a first pass I think I'm OK with this.

aspiers:
... but only if you do proper line-breaking on the `cmd` assignment!  That long line is *horrible* (it even causes a horizontal scrollbar on this PR review page).

bkutil:
@aspiers @tboerger updated the script to use Rails models instead of knife.

aspiers:
LGTM.

PR #1280, opened 2015-05-29
[WIP] Decyphering deployment queue & apply_role

bkutil:
Added comments that try to explain various bits & pieces of the deployment queue (queueing) and apply_role (deployment) algorithms we use.

Please add comments/questions/opinions.

*Do not merge*

cc: @vuntz @dirkmueller @aspiers

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -155,14 +155,19 @@ class ServiceObject
 #
 
   # Create map with nodes and their element list
+  # Transform ( {role => [nodes], role1 => [nodes]} hash to { node => [roles], node1 => [roles]},
+  # accounting for clusters
   def elements_to_nodes_to_roles_map(elements)
     nodes_map = {}
+
     elements.each do |role_name, nodes|
+      # Expand clusters to individual nodes
       nodes, failures = expand_nodes_for_all(nodes)
       unless failures.nil? || failures.empty?
         @logger.debug "elements_to_nodes_to_roles_map: skipping items that we failed to expand: #{failures.join(", ")}"
       end
 
+      # Add the role to node's list
       nodes.each do |node_name|
         if NodeObject.find_node_by_name(node_name).nil?
           @logger.debug "elements_to_nodes_to_roles_map: skipping deleted node #{node_name}"
@@ -190,31 +195,41 @@ class ServiceObject
     [ delay, pre_cached_nodes ]
   end
 
+  # Get a hash of {node => [roles], node1 => [roles]}
   def add_pending_elements(bc, inst, elements, queue_me, pre_cached_nodes = {})
     nodes_map = elements_to_nodes_to_roles_map(elements)
 
+    # We need to be sure that we're the only ones modifying the node records at this point.
+    # This will work for preventing changes from rails app, but not necessarily chef.
+    # Tough luck.
     f = acquire_lock "BA-LOCK"
+
+    # Delay is the list of nodes that are not ready and are needed for this deploy to run
     delay = []
     pre_cached_nodes = {}
     begin
       # Check for delays and build up cache
+      # FIXME: why?
       if queue_me
         delay = nodes_map.keys
       else
         delay, pre_cached_nodes = elements_not_ready(nodes_map.keys, pre_cached_nodes)
       end
 
-      # Add the entries to the nodes.
+      # We have all nodes we'll ever need. Mark them as applying and this proposal
+      # as the 'author'
       if delay.empty?
         nodes_map.each do |node_name, val|
           node = pre_cached_nodes[node_name]
 
-          # Nothing to delay so mark them applying.
+          # Nothing to wait for so mark them applying now.
           node.crowbar['state'] = 'applying'
           node.crowbar['state_owner'] = "#{bc}-#{inst}"
           node.save
         end
       else
+        # Update all nodes affected by this proposal deploy (elements) -> add info that this proposal
+        # will add list of roles to node's crowbar.pending hash.
         nodes_map.each do |node_name, val|
           # Make sure we have a node.
           node = pre_cached_nodes[node_name]
@@ -238,6 +253,10 @@ class ServiceObject
     [ delay, pre_cached_nodes ]
   end
 
+  # Each node keeps a list of roles (belonging to the current proposal) that are to be applied to it
+  # under crowbar.pending.barclamp-name hash.
+  # When we finish deploying and also when we dequeue the proposal, the list should be emptied.
+  # FIXME: looks like bc-inst: value should be a list, not a hash?
   def remove_pending_elements(bc, inst, elements)
     nodes_map = elements_to_nodes_to_roles_map(elements)
 
@@ -280,6 +299,10 @@ class ServiceObject
 #   dequeue_proposal - remove item from queue and clean up
 #   process_queue - see what we can execute
 #
+
+  # Receives proposal info (name, barclamp), list of nodes (elements), on which the proposal
+  # should be applied, and list of dependencies - a list of {barclamp, name/inst} hashes.
+  # It adds them to the queue, if possible.
   def queue_proposal(inst, elements, deps, bc = @bc_name)
     @logger.debug("queue proposal: enter #{inst} #{bc}")
     delay = []
@@ -306,6 +329,12 @@ class ServiceObject
       end
 
       # Make sure the deps if we aren't being queued.
+      # If queue_me is true, the delay contains all elements, otherwise, only nodes that are not ready.
+      # FIXME: why do we need to do this? Because we will retry/wait?
+      # FIXME: this nicely duplicates the code in process_queue
+      # * any of the dependencies does not exist
+      # * any of the dependencies is already queued
+      # * any of the dependencies failed to apply or was not applied at all
       queue_me = false
 
       deps.each do |dep|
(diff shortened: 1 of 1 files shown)


crowbar_framework/lib/deployment_queue.rb
```diff
@@ -0,0 +1,406 @@
...
+        db["proposal_queue"] << { "barclamp" => bc, "inst" => inst, "elements" => elements, "deps" => deps }
+      else
+        # Update (overwrite) item that is already in queue
+        preexisting_queued_item["elements"] = elements
+        preexisting_queued_item["deps"] = deps
+      end
+
+      db.save
+    rescue StandardError => e
+      logger.error("Error queuing proposal for #{bc}:#{inst}: #{e.message}")
```
tboerger: queueing?

crowbar_framework/lib/deployment_queue.rb
```diff
@@ -0,0 +1,406 @@
...
+  end
+
+  #
+  # NOTE: If dependencies don't form a DAG (Directed Acyclic Graph) then we have a problem
+  # with our dependency algorithm
+  #
+  def process_queue
+    logger.debug("process queue: enter")
+    loop_again = true
+    while loop_again
```
tboerger:
```
begin
end while condition
```

instead of setting the variable to true and directly to false?

crowbar_framework/lib/deployment_queue.rb
```diff
@@ -0,0 +1,406 @@
...
+        answer = service.proposal_commit(inst, true, false)
+
+        logger.debug("process queue: item #{item.inspect}: results #{answer.inspect}")
+
+        # 202 means some nodes are not ready, bail out in that case
+        # FIXME: we're re-running the whole apply continuously, until there
+        # are no items left in the queue?
+        loop_again = true if answer[0] != 202
+
+        $htdigest_reload = true
```
tboerger: If i remember correctly we don't need this anymore.

crowbar_framework/lib/deployment_queue.rb
aspiers: Apparently both are fine. http://english.stackexchange.com/questions/60852/queueing-or-queuing

bkutil:
I changed the pull so it only adds comments to the code and can be safely merged.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1205,31 +1306,40 @@ def apply_role(role, inst, in_queue)
...
+
+              # FIXME: are remove_roles a way how to tell us to do some post-cleanup?
               pending_node_actions[node_name][:add] << elem_remove unless elem_remove.nil?
+
               nodes_in_batch << node_name
             end
           end
         end
 
+        # FIXME: what does it mean if new_nodes is nil?
```
vuntz: I think that it means that we are deactivating the proposal.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1205,31 +1306,40 @@ def apply_role(role, inst, in_queue)
...
         unless new_nodes.nil?
           new_nodes.each do |node_name|
             # Don't add deleted nodes to the run order
+            # FIXME: if we want to deploy on a deleted node, shouldn't we just bail out instead?
```
vuntz: IIRC, this got added for the barclamps where all nodes are used (for instance, provisioner, logging, dns, ntp); so that we don't fail too easily when a node got forgotten.

It's kind of a ugly workaround for the fact that we don't properly handle forgotten node + for the fact that we don't have some alias that be used to assign all existing nodes to a role (which would be an improvement over the requirement to list all nodes).

crowbar_framework/app/models/service_object.rb
```diff
@@ -1238,15 +1348,21 @@ def apply_role(role, inst, in_queue)
...
     end
 
+    # Part III: Update run lists of nodes to reflect new deployment. I.e. write through
+    # the deployment schedule in pending node actions into run lists.
     @logger.debug "Clean the run_lists for #{pending_node_actions.inspect}"
+
     admin_nodes = []
+
     pending_node_actions.each do |node_name, lists|
+      # FIXME: Do we constantly need to filter out nodes that drop out during deployment?
```
vuntz: I think we've done it so far for the nodes in the new deployment, but not for the nodes in the old deployment (ie, nodes for which the proposal will be deactivated).

crowbar_framework/app/models/service_object.rb
```diff
@@ -1279,6 +1395,8 @@ def apply_role(role, inst, in_queue)
       end
 
       # Make sure the config role is on the nodes in this barclamp, otherwise remove it
+      # FIXME: All nodes is basically all new nodes, which should have an add/remove list,
+      # why do we need to add specifically the passed role?
```
vuntz: Not sure I understand the question :/

crowbar_framework/app/models/service_object.rb
```diff
@@ -1205,31 +1306,40 @@ def apply_role(role, inst, in_queue)
...
             end
 
+            # An old node that is not in the new deployment, drop it
             if new_nodes.nil? or !new_nodes.include?(node_name)
               @logger.debug "remove node #{node_name}"
               pending_node_actions[node_name] = { :remove => [], :add => [] } if pending_node_actions[node_name].nil?
+
               pending_node_actions[node_name][:remove] << role_name
+
+              # FIXME: are remove_roles a way how to tell us to do some post-cleanup?
```
vuntz: It's a way to "de-configure" things on the node when a role is not used anymore for that node. For instance, stopping a service, or removing packages.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1079,18 +1150,37 @@ def self.proposal_to_role(proposal, bc_name)
...
     @logger.debug "apply_role(#{role.name}, #{inst}, #{in_queue})"
 
+    # Part I: Looking up data & checks
+    #
+    # we look up the role in the database (if there is one), the new one
+    # is passed in as the role param.
+    #
+    # From both, we need 'elements', i.e. role -> nodes map and element_order
+    # -> an ordered list of roles, telling us in which order they should be applied.
+    # FIXME: does this give us dependency info within a barclamp?
```
vuntz: The order does provide some dependency between roles inside one barclamp, yes.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1079,18 +1150,37 @@ def self.proposal_to_role(proposal, bc_name)
...
+    #
+    # Any of the new role's elements can contain clusters, so we need to expand
+    # them to individual nodes. We store them in 'elements_expanded'.
+    # FIXME: would this work if someone reassigned/removed a node from a cluster
+    # in the old role?
```
vuntz: I think it should work, with magic in barclamp-pacemaker updating the roles when a cluster is updated.

crowbar_framework/app/models/service_object.rb
```diff
@@ -484,12 +545,22 @@ def process_queue
...
+
+        # This will call apply_role and chef-client.
+        # Params: (inst, in_queue, validate_after_save)
         answer = service.proposal_commit(inst, true, false)
+
         @logger.debug("process queue: item #{item.inspect}: results #{answer.inspect}")
+
+        # 202 means some nodes are not ready, bail out in that case
+        # FIXME: we're re-running the whole apply continuously, until there
+        # are no items left in the queue?
```
vuntz: Yes. And it's lame because from the user perspective, we're still applying the first barclamp while this part was completed and we're actually applying what was in the queue.

This needs to be fixed.

crowbar_framework/app/models/service_object.rb
```diff
@@ -313,16 +342,29 @@ def queue_proposal(inst, elements, deps, bc = @bc_name)
...
       delay, pre_cached_nodes = add_pending_elements(bc, inst, elements, queue_me)
+
+      # We have all nodes ready.
       if delay.empty?
+        # FIXME: Not sure about this, but:
+        # There's a path: process_queue -> proposal_commit -> apply_role -> queue_proposal.
+        # which seems to be used as a test if all dependencies (queue_me =
+        # false) and nodes (delay.empty?) are still in that state by the time
+        # we want to apply. So if that is the case, we just drop proposal from
+        # the queue and exit.
```
vuntz: I believe you're correct (didn't double-check, but that sounds right)

crowbar_framework/app/models/service_object.rb
bkutil: It is not clear to me, why do we need to explicitly add the role passed to the `apply_role` - wasn't it already added via `alist` a few lines above?

vuntz: Ah, okay. It's two different roles :-) Let's say you add node1 to keystone-server and apply the keystone proposal, then the lines above make sure that in chef, node1 has the keystoner-server role. The lines below, on the other hand, make sure that node1 has the keystone-config-default role (which is the role matching this specific proposal, with the attributes that are set to what is in the proposal).

So you have two sets of roles:

- the roles like keystone-server that define what recipes will be run
- the roles like keystone-config-default that define what attributes are used to configure things properly during the chef run

bkutil:
@vuntz thanks for info. I incorporated your responses/feedback to the comments.

bkutil:
@vuntz is this ok to merge?

crowbar_framework/app/models/service_object.rb
aspiers: In general there are *three* sets of roles, because there are also node roles.  This is explained by slide 65 of https://w3.suse.de/~aspiers/cloud/HA-training/HA-training-part-2a.pdf

mjura:
For me it is OK +1

tboerger:
LGTM

PR #1284, opened 2015-06-10
Fix for search_env_filtered(), (bnc#934227)

mjura:
Check if node[barclamp][:config] exist to prevent

    undefined method `[]' for nil:NilClass

It was reported in https://bugzilla.suse.com/show_bug.cgi?id=934227

### DIFF
--- a/chef/cookbooks/utils/libraries/search.rb
+++ b/chef/cookbooks/utils/libraries/search.rb
@@ -25,7 +25,7 @@ class Chef
       # There are two conventions to filter by barclamp proposal:
       #  1) Other barclamp cookbook: node[@cookbook_name][$OTHER_BC_NAME_instance]
       #  2) Same cookbook: node[@cookbook_name][:config][:environment]
-      if barclamp == cookbook_name
+      if node[barclamp] && node[barclamp][:config] && (barclamp == cookbook_name)
         env = node[barclamp][:config][:environment]
       else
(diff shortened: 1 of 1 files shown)


bkutil:
looks good, cc @vuntz

rsalevsky:
tested successfully :+1:

chef/cookbooks/utils/libraries/search.rb
```diff
@@ -25,7 +25,7 @@ def search_env_filtered(type, query="*:*", sort='X_CHEF_id_CHEF_X asc',
       # There are two conventions to filter by barclamp proposal:
       #  1) Other barclamp cookbook: node[@cookbook_name][$OTHER_BC_NAME_instance]
       #  2) Same cookbook: node[@cookbook_name][:config][:environment]
-      if barclamp == cookbook_name
+      if !node[barclamp].nil? && !node[barclamp][:config].nil? && (barclamp == cookbook_name)
```
dirkmueller:
```ruby

if node[barclamp] && node[barclamp][:config] && ...

```

would be easier to read. does that work as well?

vuntz: +1

tboerger: :+1:

mjura: Awesome, I was looking for something like this! 
@dirkmueller Thank you!

dirkmueller:
+1

PR #1286, opened 2015-06-11
[stoney] More fixes to make the "disable 15-minute chef-client run" not crash everything

vuntz:


vuntz:
cc @bkutil, tboerger: I'm not quite sure about the second commit, but `@logger` was definitely creating an issue when used from a forked process. Can you confirm my fix is correct?

dirkmueller:
LGTM

crowbar_framework/app/models/service_object.rb
```diff
@@ -1492,14 +1492,15 @@ def chef_daemon(action, node_list)
...
   def wait_for_chef_clients(node_name, options = {})
-    options = if options.fetch(:logger)
+    use_logger = options.fetch(:logger)
+    options = if use_logger
       {:logger => @logger}
     else
       {}
     end
-    @logger.debug("wait_for_chef_clients: Waiting for already running chef-clients on #{node_name}.")
+    @logger.debug("wait_for_chef_clients: Waiting for already running chef-clients on #{node_name}.") if use_logger
```
MaximilianMeister: I dont understand why this logger shouldn't work in some case

MaximilianMeister:
+1 for the first commit

crowbar_framework/app/models/service_object.rb
vuntz: That's what I mentioned in my comment on the PR: it seems `@logger` is not available after the fork. But only on ruby 1.8.

bkutil: What is the reason behind options mangling?

Wouldn't something like this work, considering `wait_for_chef_clients` is called with `options = { :logger => @logger }`?:

```ruby
logger = options.fetch(:logger) { Rails.logger }
logger.debug("wait_for...")
...
options[:logger] ||= logger
unless RemoteNode.chef_ready?(....options)
  logger.error("Waiting...")
```

vuntz: I believe the issue is that `Rails.logger` is kind of useless after the fork, and so should not be used in there. But maybe that works; unfortunately, can't try that right now.

vuntz:
I'd appreciate if people could try to fix the second issue I was fixing another way. I'm happy to take suggestions, since I was not able to find a better fix.

dirkmueller:
@MaximilianMeister , @bkutil , @rsalevsky  does it really matter to fix the 2nd commit which fixes a stoney only issue? this patch is in our packages for ages, its time to get it merged imho. 

it LGTM, but I can't review it.

crowbar_framework/app/models/service_object.rb
MaximilianMeister: Usually you use `Rails.logger` outside the models or controllers. but `Rails.logger` must always work in the full scope of the App.

MaximilianMeister:
@vuntz @dirkmueller I dont have any other suggestion. in fact i don't understand why this should be needed or what is the difference (in terms of logic). but if it makes a technical difference and if it fixes the issue then we should consider merging it.

rsalevsky:
I'm fine with merging it, the second commit is a little bit hacky but it works.

aspiers:
:-1: as it stands, sorry.

vuntz:
Obsoleted by https://github.com/crowbar/barclamp-crowbar/pull/1352

PR #1287, opened 2015-06-11
On apply_role, do not stop/start chef several times for nodes

vuntz:
If a node is on multiple roles for this proposal, then we only want to
have chef-client stopped/started once for this node.


MaximilianMeister:
+1

bkutil:
+1

PR #1288, opened 2015-06-11
Add proposal versions

bkutil:
Do not merge (the feature currently does not work due to proposal
having overrided [] method). Also, code will be separated into
smaller commits and covered by tests.

The proposal_versions table holds history of every proposal. The records are full
copies and can be used to restore the original model.

State after create, update and before destroy actions is recorded.

The idea is, that a deployment task will not apply the proposal in its current state,
but its version when 'apply' was originally clicked.

This allows the user to modify the proposal afterwards, or even delete
it, and the deployment queue will not or apply unexpected changes, or
get into an inconsistent state, as it will hold the complete info.

Additionally, UX tricks like 'undo', 'view history', 'diffing' or 'view cloud state
at past deployment', etc. are possible.

There are a few unknowns:

1. Should this 'snapshotting' extend to the proposal dependencies or
even all proposals as well? The deploy would then check that all the
dependent proposals have been applied (and do not differ), then
selectively apply them?

2. When a proposal is deleted, should we enqueue another 'cleanup'
deploy, that gives the cookbooks a chance to clean up?


tboerger:
Maybe we should simply add one versioning gem like paper_trail? That way we get versioning for all upcoming models and don't need to reinvent the wheel. If it is too generic we can extend the base functionality with an extended version model.

bkutil:
@tboerger that was the first thing I tried. The problem was, that I wasn't able to figure out how to make paper_trail work with overloaded `initialize` method, like the `Proposal` class has... as it was simply calling `new` without any parameters, it was raising an exception...

tboerger:
> @tboerger that was the first thing I tried. The problem was, that I wasn't able to figure out how to make paper_trail work with overloaded initialize method, like the Proposal class has... as it was simply calling new without any parameters, it was raising an exception...

So there have been a mismatch of the parameters for ```initialize```?

bkutil:
> So there have been a mismatch of the parameters for ```initialize```?

Yes, it raises right here:

https://github.com/airblade/paper_trail/blob/master/lib/paper_trail/version_concern.rb#L202

tboerger:
So we can not use paper_trail just because of ```raise ArgumentError.new("Barclamp attribute is required") unless attributes && attributes.key?(:barclamp)```?

bkutil:
> So we can not use paper_trail just because of ```raise ArgumentError.new("Barclamp attribute is required") unless attributes && attributes.key?(:barclamp)```?

This check is there, because the Proposal object is not valid
without a barclamp attribute.

So, that is part of the reason, yes.

The other thing is, that the observer and the acompanying model is
really simple, so I wasn't sure that pulling in a whole library into the
bundle, and having to package, maintain or even debug it wouldn't be
more work in the long term.

Finally, it seems to me (but correct me if I'm wrong), that Proposal
is the only model which really benefits from versioning, so it's OK to
write special code/table for it.

But the only hard stopper for me is lack of alternative solution for
the 'create only valid objects' issue above.

tboerger:
I'm fine with that. We hopefully doesnt need any other versioning :)

aspiers:
Looks very nice so far, great work guys! :)

rsalevsky:
@bkutil please rebase

bkutil:
@tboerger you are right, it seems that it'd be useful to have revisions for the role data as well. Lets close this for now and see, whether we could use `paper_trail` somehow.

PR #1289, opened 2015-06-18
Update ssl.conf.erb for Apache 2.4 on SLE 12 (bsc#935050)

tserong:
Apache 2.4 doesn't understand the SSLMutex directive, so without this
fix SSL radosgw (or SSL anything else) on Apache 2.4 on SLE 12 won't
work.

Signed-off-by: Tim Serong <tserong@suse.com>

### DIFF
--- a/chef/cookbooks/apache2/templates/suse/mods/ssl.conf.erb
+++ b/chef/cookbooks/apache2/templates/suse/mods/ssl.conf.erb
@@ -25,11 +25,29 @@
 	AddType application/x-x509-ca-cert .crt
 	AddType application/x-pkcs7-crl    .crl
 
+	#   SSL Cipher Suite:
+	#   List the ciphers that the client is permitted to negotiate.
+	#   See the mod_ssl documentation for a complete list.
+	#
+	SSLCipherSuite ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:ECDH+3DES:DH+3DES:RSA+AESGCM:RSA+AES:RSA+3DES:!aNULL:!MD5:!DSS
+
+	#   SSLHonorCipherOrder
+	#   If SSLHonorCipherOrder is disabled, then the client's preferences
+	#   for chosing the cipher during the TLS handshake are used.
+	#   If set to on, then the above SSLCipherSuite is used, in the order
+	#   given, with the first supported match on both ends.
+	SSLHonorCipherOrder on
+
 	#   Pass Phrase Dialog:
 	#   Configure the pass phrase gathering process.
 	#   The filtering dialog program (`builtin' is a internal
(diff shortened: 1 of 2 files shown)


tserong:
Note: I've tested this for SSL radosgw and calamari deployments on SLE 12, and it works fine. I have not tested it for SSL deployments of anything on SLE 11. IMO the config change looks clean, and I don't *think* it will cause any dramas, but thinking things does not always make them so...

tboerger:
LGTM

rsalevsky:
+1 did some basic testing

vuntz:
@rsalevsky just to clarify: did you test with horizon ssl? On SLE11 or SLE12?

chef/cookbooks/apache2/templates/suse/mods/ssl.conf.erb
```diff
@@ -38,17 +56,15 @@
...
-	SSLSessionCache         shmcb:/var/lib/apache2/ssl_scache(512000)
-	SSLSessionCacheTimeout  600
+	#</IfModule>
 
-	#   This configures the SSL engine's semaphore (aka. lock) which is
-	#   used for mutual exclusion of operations which have to be done in a
-	#   synchronized way between the pre-forked Apache server processes.
-	#   "default" tells the SSL Module to pick the default locking
-	#   implementation as determined by the platform and APR.
-	SSLMutex  default
```
vuntz: Should we keep that if apache version is <= 2.2? (default is "none", not "default")

rsalevsky:
@vuntz SLE11, Horizion is not allowed on SLE12

vuntz:
> @vuntz SLE11, Horizion is not allowed on SLE12

It is in master / Cloud 6 ;-)

chef/cookbooks/apache2/templates/suse/mods/ssl.conf.erb
tserong: Dunno.  If we do, it needs to go inside <IfVersion>, which means an `a2enmod version` somewhere.

tserong: Hrm, actually, looking at http://httpd.apache.org/docs/2.2/mod/mod_ssl.html#sslmutex it says "none: ... Use at own risk", due to possibility of sometimes mangling a session cache. I guess I should see about putting it back in for older Apache versions.

tserong:
OK, I've added an `<IfVersion>` guard around the SSLMutex directive. If that looks OK, can someone with a SLE 11 deploy do a quick test, just to assuage my paranoia please? ;-)

vuntz:
+1 but would also welcome some quick test on SLE11 :-)

tserong:
Ping anyone with a handy SLE 11 test rig...? ;-)

dirkmueller:
+1

vuntz:
Tested on SLE11, works.

PR #1290, opened 2015-06-18
Add validation module with method for package name checking

thutterer:
After some discussion in crowbar/barclamp-provisioner#427 @aspiers suggested it would be best to have a module for validations like the RegEx from mentioned Pull Request.
This way it's easy to write tests for these validations.

This PR adds such a module, a method for package name validation and its test.

### DIFF
--- a/crowbar_framework/lib/crowbar/validator/package_name_validator.rb
+++ b/crowbar_framework/lib/crowbar/validator/package_name_validator.rb
@@ -1,0 +1,11 @@
+module Crowbar
+  module Validator
+    class PackageNameValidator
+
+      def validate(package_name)
+        package_name =~ /\A(\w[\w\+\.-]+(?:[<>]?=?[\w\+\.-]+)?)\z/
+      end
+
+    end
+  end
+end
--- a/crowbar_framework/spec/lib/crowbar/validator/package_name_validator_spec.rb
+++ b/crowbar_framework/spec/lib/crowbar/validator/package_name_validator_spec.rb
@@ -1,0 +1,44 @@
+require 'spec_helper'
+
+describe Crowbar::Validator::PackageNameValidator do
+  it "allows valid package names" do
+    valid_names = [
+      "letters",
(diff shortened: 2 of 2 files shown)


tboerger:
Wrong path, it should be conventional ```lib/crowbar/validation.rb```

aspiers:
@tboerger I don't disagree, but curious why it should be `lib/crowbar/` ?

aspiers:
Oh, to match `module Crowbar`, I guess?

thutterer:
So I move that file into `lib/crowbar` and rename the module+file to `Validator` instead of `Validation`, correct?

thutterer:
Updated with what I think @tboerger suggested

tboerger:
> @tboerger I don't disagree, but curious why it should be lib/crowbar/ ?

Yeah, to match the namespace.

tboerger:
:+1:

aspiers:
:+1: Awesome, thanks :)

jsuchome:
+1

bkutil:
Looks good as it is (+1), but it would get an extra star if it followed the rails validators 'api': http://guides.rubyonrails.org/active_record_validations.html#custom-validators.

i.e. to be in: `lib/crowbar/validator/package_name_validator.rb` file and
```
  module Crowbar
    module Validator
     class PackageNameValidator
```
with `validate` being an instance method. This way, all our validators will have the same structure and reside in the same directory, which will be less confusing and somewhat more convenient.

tboerger:
Valid and good suggestion from @bkutil

thutterer:
> Valid and good suggestion from @bkutil

Think so too. I will adopt that.
Is the `spec.rb` then still in the right place?

tboerger:
The spec have to reflect the path of the lib :)

thutterer:
Updated with the suggestions from @tboerger and @bkutil

bkutil:
still +1 :)

PR #1292, opened 2015-06-24
Fixing Reinstall, Power Off and Power Cycle for the windows nodes

ovisan:
https://trello.com/c/ZnrdM70i/382-3-p10-make-reinstall-work-for-windows-nodes

### DIFF
--- a/chef/cookbooks/crowbar/recipes/default.rb
+++ b/chef/cookbooks/crowbar/recipes/default.rb
@@ -36,6 +36,9 @@ when "ubuntu","debian"
     libshadow-ruby1.8
     markdown
   )
+  unless search(:node, "platform:windows").empty?
+    pkglist.push "smbclient"
+  end
 
   rainbows_path = "/var/lib/gems/1.8/bin/"
 when "redhat","centos"
@@ -46,6 +49,9 @@ when "redhat","centos"
     sqlite-devel
     python-markdown
   )
+  unless search(:node, "platform:windows").empty?
+    pkglist.push "samba-client"
+  end
 
   rainbows_path = ""
 when "suse"
@@ -78,6 +84,9 @@ when "suse"
     ruby2.1-rubygem-sqlite3
     ruby2.1-rubygem-syslogger
   )
+  unless search(:node, "platform:windows").empty?
+    pkglist.push "samba-client"
+  end
 end
 
 pkglist.each do |p|
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -1138,6 +1138,30 @@ class NodeObject < ChefObject
     end
   end
 
+  def net_rpc_cmd(cmd)
+    case cmd
+    when :power_cycle
+      unless system("net", "rpc", "shutdown", "-f", "-r", "-I", @node.name ,"-U", "Administrator%#{@node[:provisioner][:windows][:admin_password]}")
+        Rails.logger.warn("samba command \"#{cmd}\" for #{@node.name} failed - node in unknown state")
+      end
+    when :power_off
+      unless system("net", "rpc", "shutdown", "-f", "-I", @node.name ,"-U", "Administrator%#{@node[:provisioner][:windows][:admin_password]}")
+        Rails.logger.warn("samba command \"#{cmd}\" for #{@node.name} failed - node in unknown state")
+      end
+    when :reboot
+      unless system("net", "rpc", "shutdown", "-r", "-I", @node.name ,"-U", "Administrator%#{@node[:provisioner][:windows][:admin_password]}")
+        Rails.logger.warn("samba command \"#{cmd}\" for #{@node.name} failed - node in unknown state")
+      end
+    when :shutdown
+      unless system("net", "rpc", "shutdown", "-I", @node.name ,"-U", "Administrator%#{@node[:provisioner][:windows][:admin_password]}")
+        Rails.logger.warn("samba command \"#{cmd}\" for #{@node.name} failed - node in unknown state")
+      end
+    else
+      Rails.logger.warn("samba command #{cmd} failed for #{@node.name}.")
+    end
+  end
+
+
   def bmc_cmd(cmd)
     if bmc_address.nil? || get_bmc_user.nil? || get_bmc_password.nil? ||
         !system("ipmitool", "-I", "lanplus", "-H", bmc_address, "-U", get_bmc_user, "-P", get_bmc_password, cmd)
@@ -1210,8 +1234,11 @@ class NodeObject < ChefObject
       rescue Timeout::Error
         Rails.logger.warn("chef client seems to be still running after 5 minutes of wait; going on with the reboot")
       end
-
-      ssh_cmd("/sbin/reboot")
+      if @node[:platform] == "windows"
+        net_rpc_cmd(:power_cycle)
+      else
+        ssh_cmd("/sbin/reboot")
+      end
     end
     results
   end
@@ -1234,12 +1261,20 @@ class NodeObject < ChefObject
 
   def reboot
     set_state("reboot")
-    ssh_cmd("/sbin/reboot")
+    if @node[:platform] == "windows"
+      net_rpc_cmd(:reboot)
+    else
+      ssh_cmd("/sbin/reboot")
+    end
   end
 
   def shutdown
     set_state("shutdown")
-    ssh_cmd("/sbin/poweroff")
+    if @node[:platform] == "windows"
+      net_rpc_cmd(:shutdown)
+    else
+      ssh_cmd("/sbin/poweroff")
+    end
   end
 
   def poweron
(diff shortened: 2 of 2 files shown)


crowbar_framework/app/models/node_object.rb
```diff
@@ -1210,8 +1224,11 @@ def set_state(state)
       rescue Timeout::Error
         Rails.logger.warn("chef client seems to be still running after 5 minutes of wait; going on with the reboot")
       end
-
-      ssh_cmd("/sbin/reboot")
+      if @node[:platform] == "windows"
+        bmc_cmd("power cycle")
```
bkutil: Instead of adding windows related functionality to `bmc_cmd` method, would it make sense to make make a similar, windows-specific method, say `net_rpc_cmd` and use it here? Imho it'd be cleaner to separate the per-platform functionality into different methods.

tboerger: :+1:

ovisan: @bkutil that would require extra checks for "power cycle" and "power off" because ssh_cmd will not longer apply, so having it in bmc is better.

tboerger: But it still makes sense to add a ```net_rpc_cmd``` method to avoid the code duplication.

vuntz: Agreed.

bkutil: @ovisan yes, there would be a similar `case` as in `bmc_cmd`, but there would not be any checks for platform.

crowbar_framework/app/models/node_object.rb
```diff
@@ -1138,6 +1138,25 @@ def ssh_cmd(cmd)
     end
   end
 
+  def net_rpc_cmd(cmd)
+    case cmd
+    when "power cycle"
```
tboerger:
```
:power_cycle
```

crowbar_framework/app/models/node_object.rb
```diff
@@ -1138,6 +1138,25 @@ def ssh_cmd(cmd)
...
+      unless system("net", "rpc", "shutdown", "-f", "-r", "-I", @node.name ,"-U", "Administrator%#{@node[:provisioner][:windows][:admin_password]}")
+        Rails.logger.warn("samba command \"#{cmd}\" for #{@node.name} failed - node in unknown state")
+        return nil
+      end
+    when "power off"
```
tboerger:
```
:power_off
```

crowbar_framework/app/models/node_object.rb
```diff
@@ -1210,8 +1229,11 @@ def set_state(state)
       rescue Timeout::Error
         Rails.logger.warn("chef client seems to be still running after 5 minutes of wait; going on with the reboot")
       end
-
-      ssh_cmd("/sbin/reboot")
+      if @node[:platform] == "windows"
+        net_rpc_cmd("power cycle")
```
tboerger: Change to the above symbol

crowbar_framework/app/models/node_object.rb
```diff
@@ -1234,12 +1256,20 @@ def reset
 
   def reboot
     set_state("reboot")
-    ssh_cmd("/sbin/reboot")
+    if @node[:platform] == "windows"
+      net_rpc_cmd("power cycle")
```
tboerger: Change to the above symbol

crowbar_framework/app/models/node_object.rb
```diff
@@ -1234,12 +1256,20 @@ def reset
...
+    else
+      ssh_cmd("/sbin/reboot")
+    end
   end
 
   def shutdown
     set_state("shutdown")
-    ssh_cmd("/sbin/poweroff")
+    if @node[:platform] == "windows"
+      net_rpc_cmd("power off")
```
tboerger: Change to the above symbol

crowbar_framework/app/models/node_object.rb
```diff
@@ -1249,12 +1279,20 @@ def poweron
 
   def powercycle
     set_state("reboot")
-    bmc_cmd("power cycle")
+    if @node[:platform] == "windows"
+      net_rpc_cmd("power cycle")
```
tboerger: Change to the above symbol

crowbar_framework/app/models/node_object.rb
```diff
@@ -1249,12 +1279,20 @@ def poweron
...
+    else
+      bmc_cmd("power cycle")
+    end
   end
 
   def poweroff
     set_state("shutdown")
-    bmc_cmd("power off")
+    if @node[:platform] == "windows"
+      net_rpc_cmd("power off")
```
tboerger: Change to the above symbol

crowbar_framework/app/models/node_object.rb
```diff
@@ -1234,12 +1256,20 @@ def reset
 
   def reboot
     set_state("reboot")
-    ssh_cmd("/sbin/reboot")
+    if @node[:platform] == "windows"
+	    net_rpc_cmd(:power_cycle)
```
tboerger: Too much indention :)

tboerger:
I dislike the code duplication a little bit but :+1:

crowbar_framework/app/models/node_object.rb
```diff
@@ -1138,6 +1138,22 @@ def ssh_cmd(cmd)
     end
   end
 
+  def net_rpc_cmd(cmd)
+    case cmd
+    when :power_cycle
+      unless system("net", "rpc", "shutdown", "-f", "-r", "-I", @node.name ,"-U", "Administrator%#{@node[:provisioner][:windows][:admin_password]}")
```
vuntz: Shouldn't we make sure somehow that the samba client tools are installed? (probably in the crowbar cookbook)

crowbar_framework/app/models/node_object.rb
```diff
@@ -1249,12 +1276,20 @@ def poweron
 
   def powercycle
     set_state("reboot")
-    bmc_cmd("power cycle")
+    if @node[:platform] == "windows"
+      net_rpc_cmd(:power_cycle)
```
vuntz: The calls for `reboot` and `powercycle` should be slightly different (at least, they are for non-windows nodes: `reboot` vs `reboot -f`). Do I understand right that the `-f` parameter to `net` is also used to force a shutdown/reboot? In that case, we probably don't want to do it in `reboot`/`shutdown`, but only in `powercycle`/`poweroff`.

crowbar_framework/app/models/node_object.rb
ovisan: samba is installed if windows node is deployed, because to access the tftpboot directory in linux, the samba share is created and accessed from windows.

crowbar_framework/app/models/node_object.rb
ovisan: From net documentation:
-f
Force shutting down all applications.

It can be executed also without -f, but to force dangling applications to close is safer.
in both  cases power cycle and reboot, poweroff shutdown respectively, the operation is the same, that's why it's submitted like this.

crowbar_framework/app/models/node_object.rb
vuntz: That's correct, but if you kill your admin server and restore it for backup, then you might not have samba installed anymore. Also, this is the server side -- not sure this would automatically install the client side.

crowbar_framework/app/models/node_object.rb
vuntz: > It can be executed also without -f, but to force dangling applications to close is safer.

I think that's partly why we have the two different methods. A reboot is supposed to be clean, so without `-f`.

ovisan: @vuntz  added reboot and shutdown without the force switch

vuntz: Thanks!

crowbar_framework/app/models/node_object.rb
tboerger: Ok, so @ovisan have to add the package to https://github.com/crowbar/barclamp-crowbar/blob/master/chef/cookbooks/crowbar/recipes/default.rb#L29-L81 but just in case if there is any windows nodewithin crowbar available. @vuntz do you have a hint for @ovisan how he can check if we have any windows node? :)

vuntz: Search for nodes that have windows as platform?

ovisan: I need the function name, how can i search all the nodes if they have windows?

vuntz: Something like `search(:node, "platform:windows")`. You can find plenty of search examples in the cookbooks we already have...

ovisan: @vuntz added checking for the windows nodes and installing samba client only if there are windows nodes present

vuntz:
@tboerger: if you're fine installing the package in this recipe, then +1 from me.

jsuchome:
Looks good to me. +1

chef/cookbooks/crowbar/recipes/default.rb
```diff
@@ -36,6 +36,9 @@
     libshadow-ruby1.8
     markdown
   )
+  unless search(:node, "platform:windows").empty?
+    pkglist.push "smbclient"
```
aplanas: Not a Ubuntu or Debian user but at least in Debian the name is `samba-client`

https://wiki.debian.org/SambaServerSimple

ovisan: http://packages.ubuntu.com/trusty/net/smbclient

Well, the package name is smbclient.

tboerger: https://packages.debian.org/jessie/smbclient ```smbclient``` is correct for Debian as well.

tboerger:
:+1:

aplanas:
LGTM +1 (but not tested)

PR #1293, opened 2015-06-24
Handle /nodes/families json requests properly (bsc#926395)

rsalevsky:
Handle json request properly.

### DIFF
--- a/crowbar_framework/app/controllers/nodes_controller.rb
+++ b/crowbar_framework/app/controllers/nodes_controller.rb
@@ -219,6 +219,11 @@ class NodesController < ApplicationController
         })
       end
     end
+
+    #UI-only method
+    respond_to do |format|
+      format.html
+    end
   end
 
   def group_change
--- a/crowbar_framework/spec/controllers/nodes_controller_spec.rb
+++ b/crowbar_framework/spec/controllers/nodes_controller_spec.rb
@@ -154,6 +154,12 @@ describe NodesController do
       get :families
       assigns(:families).keys.should include(node.family.to_s)
     end
+
+    it "as json is not acceptable" do
+      expect {
+        get :families, :format => "json"
+      }.to raise_error
+    end
   end
 
   describe "GET status" do


aspiers:
Please make the commit message explain the change, not just refer to a bugzilla entry (which is probably only accessible by SUSE employees).  Also it is best practice to include the full URL to the bugzilla entry in the body of the commit message.  (This is [documented internally](https://etherpad.nue.suse.com/p/cloud-pull-requests) BTW.)

Here's an example of the ideal commit message: https://github.com/crowbar/barclamp-pacemaker/commit/ad8c04e5f6f0daf4c896b8ba54d96745cda3dd5d

rsalevsky:
@aspiers You are to fast. :P Was just a test to see how travis handles the unit tests.

aspiers:
Hah :)

crowbar_framework/app/controllers/nodes_controller.rb
```diff
@@ -219,6 +219,10 @@ def families
         })
       end
     end
+
+    respond_to do |format|
+      format.html { render "families" }
```
tboerger: Why? You can skip the render. Conventional you want to write just ```format.html``` and nothing more in that case.

crowbar_framework/spec/controllers/nodes_controller_spec.rb
```diff
@@ -154,6 +154,11 @@
       get :families
       assigns(:families).keys.should include(node.family.to_s)
     end
+
+    it "as json is not acceptable" do
```
tboerger: IMHO not a really useful spec. We don't really care if it responds to JSON or something different as long as a proper page gets displayed.

bkutil:
@aspiers @tboerger the pull is marked as WIP, so give the guy some space to experiment :) You can always shoot it to pieces with your comments later :)

tboerger:
> @aspiers @tboerger the pull is marked as WIP, so give the guy some space to experiment :) You can always shoot the patch to pieces with your comments later :)

I don't want to shoot his patch in pieces. I just want to give early feedback to improve his stuff :)

crowbar_framework/spec/controllers/nodes_controller_spec.rb
rsalevsky: So you mean we can skip this test?

tboerger: I think so. @bkutil what do you say?

dirkmueller: are we really arguing against a regression test here?

tboerger: > I think we're generally preferring TDD

Sure, but useful tests.

dirkmueller: it was a bugreport by QA, it makes sense to add a test that checks that there is not again a 500. plus its a nice learning excercise on how to write specs..

aspiers:
@bkutil It wasn't marked WIP when I reviewed it ;-)  But yes like @tboerger I'm just trying to be helpful ;-)

crowbar_framework/spec/controllers/nodes_controller_spec.rb
tboerger: This spec doesn't work anyway. Travis fails because of that.

bkutil:
@rsalevsky - I'm not 100% sure, but I think @vmoravec 's bug report was about the action not being able to handle requests for `JSON` If so, a proper fix would be then:

1. add `render json: @families` to the controller
2. add a trivial spec checking that response is success and it contains something meaningful.

tboerger:
@bkutil The report was that he get a 500 error page when he tries to access it with JSON format. Instead of a 500 it should simply print a 404 error and nothing more. IMHO we don't need the family stuff in JSON format.

vuntz:
I guess the question is whether this is part of the API or not. If it is, then json should be an option. If not, then 404 is fine.

tboerger:
> I guess the question is whether this is part of the API or not. If it is, then json should be an option. If not, then 404 is fine.

Currently we are using barely the HTML output, but not the JSON output. So from my side a 404 should be fine.

rsalevsky:
Ready for review

tboerger:
:+1:

crowbar_framework/app/controllers/nodes_controller.rb
```diff
@@ -219,6 +219,10 @@ def families
         })
       end
     end
+
+    respond_to do |format|
+      format.html
+    end
```
vuntz: Can we get a comment at the beginning of the method to say that this is a UI-only method, not an API method?

dirkmueller:
+1

PR #1297, opened 2015-06-30
Fix display_name when barclamps[barclamp] doesn't exists (bsc#935233)

rsalevsky:
`barclamps[barclamp]` is not always available (e.g. `barclamp-deployer`) in this
case we also want the barclamp name.

### DIFF
--- a/crowbar_framework/lib/barclamp_catalog.rb
+++ b/crowbar_framework/lib/barclamp_catalog.rb
@@ -64,12 +64,13 @@ class BarclampCatalog
   end
 
   def display_name(barclamp)
-    display = barclamps[barclamp]['display']
-
-    if display.nil? or display.empty?
-      barclamp.titlecase
+    if barclamps[barclamp] && !barclamps[barclamp]["display"].blank?
+      barclamps[barclamp]["display"]
     else
-      display
+      Rails.logger.warn(
+        "Could not find barclamp #{barclamp} in the crowbar catalog."
+      )
+      barclamp.titleize
     end
   end
 


crowbar_framework/lib/barclamp_catalog.rb
```diff
@@ -64,12 +64,14 @@ def category(barclamp)
...
-      barclamp.titlecase
+    if barclamps[barclamp]
+      if barclamps[barclamp]["display"].blank?
+        barclamp.titlecase
+      else
+        barclamps[barclamp]["display"]
+      end
     else
-      display
+      barclamp.titlecase
```
tboerger: ```Alias for String#titleize```, take directly ```titleize```

tboerger:
:+1:

vuntz:
How is it possible that a barclamp is not in the catalog? That sounds like the real bug.

tboerger:
> How is it possible that a barclamp is not in the catalog? That sounds like the real bug.

Maybe this is just related for development environments before the barclamp_install script have been executed.

tboerger:
@rsalevsky I think we want to provide a real link to bugzilla in the commit messages, not only the bug number.

rsalevsky:
@vuntz This can also happen when a barclamp is not installed correctly. The patch fixes the crowbar crash but not the bug itself.

vuntz:
> @vuntz This can also happen when a barclamp is not installed correctly.

And how is this possible? I'm pretty sure that there will be tons of other issues when that happens, so I don't think fixing this crash helps because the next step would be to fix the following crash or bug that results from this broken condition.

vuntz:
That being said, I'm not opposed to merging this, but I don't think we should consider the bug fixed.

tboerger:
The bug itself is not fixed. But we keep the UI intact. Maybe adding some kind of log message should still help.

crowbar_framework/lib/barclamp_catalog.rb
```diff
@@ -64,12 +64,14 @@ def category(barclamp)
   end
 
   def display_name(barclamp)
-    display = barclamps[barclamp]['display']
+    Rails.logger.warn(
```
tboerger: If you put it into the else than you can skip the condition.

dirkmueller:
+1

PR #1298, opened 2015-06-30
Update "Provided by" text and links

rsalevsky:
The default crowbar should be provider neutral as possible. So let's
move the standard to the crowbar website.


tboerger:
:+1:
But don' forget to rebase the suse ui branch :)

PR #1299, opened 2015-07-01
Fix 'unique_device_for()' for SLES12 virtio devices (bsc#936712)

rhafer:
In some setups on SLES12 there are no /dev/disk-by*/ links at all for virtio
disks. Return /dev/vdX as the unique name immediately in such cases.


mjura:
LGTM but I'm going to test it

mjura:
Works like a charm +1 

Please backport this change to tex

rsalevsky:
+1

PR #1302, opened 2015-07-03
Relax rubocop config

rsalevsky:


tboerger:
:+1:

tboerger:
Just to show some stats, rubocop currently complains about ```5366``` just in this barclamp!
Here is the list of offenses: https://gist.githubusercontent.com/anonymous/2ca65597cf68f45b5ac4/raw/rubocop.txt

MaximilianMeister:
Generally I am still not sure if this is really beneficial right now. looking at the stats and number of offenses. IMO think it does not integrate well at this point in time

tboerger:
> Generally I am still not sure if this is really beneficial right now. looking at the stats and number of offenses. IMO think it does not integrate well at this point in time

If not yet we will never do it.

MaximilianMeister:
you can merge this i am fine with that. +1
we should only test it a bit further to see how it integrates with the current code base.
my fear was only that we have to work around or ignore a lot of hound comments, but maybe adjusting the constraints helps to find a compromise.

PR #1303, opened 2015-07-06
[tex] Handle /nodes/families json requests properly (bsc#926395)

rsalevsky:
We do not handle json request for this page. Let's return a prober error
code instead of run in a "internal server error".

(cherry picked from commit 17734a7ef8a46738b8d61ced92d2517187bb80e4)


dirkmueller:
+1

PR #1305, opened 2015-07-07
Minor cleanups in crowbar cookbook + add crowbar.service systemd file for SLE12

vuntz:


chef/cookbooks/crowbar/recipes/default.rb
```diff
@@ -352,23 +339,60 @@
...
+      mode "0644"
+      action :create
+      source "crowbar.service"
+    end
+
+    # Make sure that any dependency change is taken into account
+    bash "reload systemd after crowbar update" do
+      code "systemctl daemon-reload"
+      action :nothing
+      subscribes :run, resources("cookbook_file[/etc/systemd/system/crowbar.service]"), :immediately
```
vuntz: @tboerger what would be the proper way of writing this line in multiline? Just go back to a new line after each `,`?

tboerger: Yes, something like that.

tboerger:
Some tiny Hound offense but LGTM

rsalevsky:
@tboerger The new SUSE style says Metrics/LineLength Max 100.

:+1:

chef/cookbooks/crowbar/recipes/default.rb
```diff
@@ -352,23 +339,60 @@
...
-    action :create
-    to "/etc/init.d/crowbar"
-    not_if "test -L /usr/sbin/rccrowbar"
+    link "/usr/sbin/rccrowbar" do
+      action :create
+      to "/etc/init.d/crowbar"
+    end
+
+    # Make sure that any dependency change is taken into account
+    bash "insserv crowbar service" do
```
tboerger: This is already in service enable implied?

vuntz: Yes when the service is not enabled, but if it's already enabled and the init file changes, then we need to call insserv to update the dependency chain.

tboerger: Ok

tboerger:
It's not beautiful but it's the case for that cookbook anyway :D
:+1:

PR #1306, opened 2015-07-08
Enable defining roles for node's disks

jsuchome:
This is the initial attempt to implement selection of disk roles (see https://trello.com/c/VaemNu77/407-5-p8-crowbar-add-ability-to-assign-intended-role-to-disks-of-a-node-to-help-with-claiming).

When such role (ceph,swift, sbd etc.) exists, other barclamps should use it when trying to claim the disks.

EDIT: See screenshot at http://imagebin.suse.de/1861/img . Items that are not editable are already claimed, the rest can still be changed.


crowbar_framework/app/models/node_object.rb
```diff
@@ -256,6 +256,27 @@ def intended_role=(value)
     @node["crowbar_wall"]["intended_role"] = value
   end
 
+  def disks_roles=(value)
+    @node["crowbar_wall"] ||= {}
+    @node["crowbar_wall"]["disks_roles"] = value
+  end
+
+  def disks_roles
+    disks_roles = crowbar_wall["disks_roles"] || {}
```
vuntz: I know it's WIP, but just one question: is it intended that you're not using `crowbar_wall["claimed_disks"]`?

crowbar_framework/app/views/nodes/_form.html.haml
```diff
@@ -113,6 +113,25 @@
...
+              = t(".disk_name")
+            %th.col-sm-2
+              = t(".disk_role")
+        %tbody
+          - @node.disks_roles.each do |disk_name, d|
+            %tr
+              %td
+                = text_field_tag "disks_roles[#{disk_name}][name]", disk_name, :class => "form-control"
+              %td
+                = text_field_tag "disks_roles[#{disk_name}][role]", d["role"], :placeholder => t(".disk_role_undecided"), :class => "form-control"
```
vuntz: I'd feel more comfortable if that'd be a menu tag so that people can't enter something wrong. Would require a way for barclamps to say they can claim disks, though.

crowbar_framework/app/views/nodes/_form.html.haml
```diff
@@ -113,6 +113,25 @@
...
+          %tr
+            %th.col-sm-3
+              = t(".disk_name")
+            %th.col-sm-2
+              = t(".disk_role")
+        %tbody
+          - @node.disks_roles.each do |disk_name, d|
+            %tr
+              %td
+                = text_field_tag "disks_roles[#{disk_name}][name]", disk_name, :class => "form-control"
```
vuntz: Should this really be editable?

jsuchome: No, sure.

crowbar_framework/app/views/nodes/_form.html.haml
jsuchome: Yes, I thought about it. Or we might add a static list

crowbar_framework/app/models/node_object.rb
jsuchome: Should I? I'm using ``unclaimed_physical_drives`` and only offering them for editing. 

Than, yes, claimed disks should be shown as well, but I didn't get that far

jsuchome: Now I also show claimed disks in the UI

rsalevsky:
When the unnecessary spacing is fixed I'm fine with that.

jsuchome:
@rsalevsky that spacing might make sense with the new code, but look at the whole map there, the spacing is done to align the items for better readability. I'm fine with changing that, but I think it doesn't make much sense....

crowbar_framework/app/controllers/nodes_controller.rb
```diff
@@ -399,6 +404,7 @@ def save_node
         :description   => :description,
         :availability_zone => :availability_zone,
         :intended_role => :intended_role,
+        :disks_roles   => :disks_roles,
```
rsalevsky: Let's ignore this for now.

rsalevsky:
+1

crowbar_framework/app/models/node_object.rb
jsuchome: So, alternatively, I could directly claim selected disks in ``node_object.rb``. But than I thought claimed disks should not be editable...

crowbar_framework/app/models/node_object.rb
```diff
@@ -256,6 +256,32 @@ def intended_role=(value)
...
+    @node["crowbar_wall"] ||= {}
+    @node["crowbar_wall"]["disk_roles"] = value
+  end
+
+  def disk_roles
+    disk_roles = crowbar_wall["disk_roles"] || {}
+
+    # remove devices that were claimed after they role was set
+    disk_roles.each do |d, _role|
+      crowbar_wall["disk_roles"].delete d if crowbar_wall[:claimed_disks].key? d
```
bkutil: Would you mind adding `()`'s around the method arguments, please? Makes the stuff a bit easier to read.

crowbar_framework/app/models/node_object.rb
```diff
@@ -256,6 +256,32 @@ def intended_role=(value)
...
+    end
```
bkutil: Instead of using `disk_roles.each` + `delete`, it should be possible to use `reject` on the `crowbar_wall` directly, I think. Same below on line 277. More generally, is the `disk_roles` helper variable needed at all?

crowbar_framework/app/models/node_object.rb
```diff
@@ -256,6 +256,32 @@ def intended_role=(value)
...
+    # disk_roles hash if it is not there yet.
+    # (New disk could have been added to the node)
+    unclaimed_physical_drives.each do |d, _data|
+      unique_name = unique_device_for(d)
+      unless disk_roles[unique_name]
+        crowbar_wall["disk_roles"] ||= {}
+        crowbar_wall["disk_roles"][unique_name] = ""
+      end
+    end
+    crowbar_wall["disk_roles"] || {}
```
bkutil: The 'side-effecting' setter method is necessary, because the disks can be out of sync when we need the list IIUC. But maybe it'd read a bit better if the 'sync' operations would be in separate methods, e.g.:

```ruby
def disk_roles
   drop_claimed_disks
   add_unclaimed_disks

   crowbar_wall["disk_roles"]
end
```

jsuchome: Hm, maybe. But those functions would be called only from this place and are only usable for disk_roles specific operations, wouldn't it be confusing to have them outside of this block?

bkutil: They could be just 'grouped' together in the code. Or maybe within a separate module (say, DiskHelpers, with these methods 'private'), included into NodeObject. But whatever works :)

jsuchome:
@bkutil reject somehow doesn't work on those items, so I've splited the code to functions at least

crowbar_framework/app/models/node_object.rb
```diff
@@ -256,6 +256,39 @@ def intended_role=(value)
...
 
+  def disk_roles=(value)
+    @node["crowbar_wall"] ||= {}
+    @node["crowbar_wall"]["disk_roles"] = value
+  end
+
+  # remove devices that were claimed after they role was set
+  def drop_claimed_disks_from_roles
+    crowbar_wall["disk_roles"].each do |name, _role|
+      if crowbar_wall[:claimed_disks].key? (name)
```
tboerger: I think ```crowbar_wall[:claimed_disks][name]``` is more intentional.

crowbar_framework/app/models/node_object.rb
```diff
@@ -256,6 +256,39 @@ def intended_role=(value)
...
+        crowbar_wall["disk_roles"].delete (name)
```
tboerger: No space in front of ```()```

crowbar_framework/app/models/node_object.rb
```diff
@@ -256,6 +256,39 @@ def intended_role=(value)
...
+      end
+    end
+  end
+
+  def add_unclaimed_disks_to_roles
+    # Check for current unclaimed disks, and add unclaimed disk to
+    # disk_roles hash if it is not there yet.
+    # (New disk could have been added to the node)
+    unclaimed_physical_drives.each do |d, _data|
```
tboerger: As you are interested only in the keys just write ```unclaimed_physical_drives.keys.each do |d|```

tboerger:
Some smaller fixes, after that I'm fine with that. Beside the fact that Travis is failing.

crowbar_framework/app/models/node_object.rb
jsuchome: ouch! just realized it is already an Array...

jsuchome:
Regarding houndci rants: that's just the old code indented to one more level, I wouldn't change style there.

crowbar_framework/app/helpers/nodes_helper.rb
```diff
@@ -88,16 +88,22 @@ def nodes_hash(group = nil)
...
+              :handle => node.handle,
+              :alias => node.alias,
+              :title => node.description(false, true),
+              :admin => node.admin?,
+              :group => node.group,
+              :platform => node[:platform],
+              :platform_version => node[:platform_version],
+              :cluster => false
+            }
+            if node["crowbar_wall"] && node["crowbar_wall"]["disk_roles"]
```
tboerger: Why don't you just add ```disk_roles => node.disk_roles``` to the hash?

tboerger:
> Regarding houndci rants: that's just the old code indented to one more level, I wouldn't change style there.

If you just add this entry to the hash like i suggested this changes won't popup :)

crowbar_framework/app/helpers/nodes_helper.rb
jsuchome: Well, it doesn't need to exist there, that's the reason for the checks. Or do you mean anything else?

tboerger: Why does it not exist? It's iterating over the NodeObject, so this method have to exist and would result in:

```ruby
        NodeObject.all.each do |node|
          nodes[node.name] = {
            :handle => node.handle,
            :alias => node.alias,
            :title => node.description(false, true),
            :admin => node.admin?,
            :group => node.group,
            :platform => node[:platform],
            :platform_version => node[:platform_version],
            :cluster => false,
            :disk_roles => node.disk_roles
          } if node.group == group or group.nil?
```

jsuchome: OK, but at this place I didn't want to call the method (so there's no iteration over disks when generating this hash), but only use the data already saved in ["crowbar_wall"]["disk_roles"]

jsuchome:
>> Regarding houndci rants: that's just the old code indented to one more level, I wouldn't change style there.

> If you just add this entry to the hash like i suggested this changes won't popup :)

``Style/HashSyntax: Use the new Ruby 1.9 hash syntax.`` does not seem to be related to new addition of disk_roles

tboerger:
> Style/HashSyntax: Use the new Ruby 1.9 hash syntax. does not seem to be related to new addition of disk_roles

I think we can ignore this offense for now, this gets auto corrected on the repo merge.

crowbar_framework/app/helpers/nodes_helper.rb
tboerger: But why not? It's not executing commands, it's just mangling a hash :)

jsuchome: Well, several hashes by several functions, starting with
https://github.com/jsuchome/barclamp-crowbar/blob/disks_roles/crowbar_framework/app/models/node_object.rb#L274

tboerger: Yeah, but mangling hashes is pretty fast :)

jsuchome: Including ``unique_device_for`` ... calling methods and other methods... maybe it's fast, but why do it when I don't have to? If ``["crowbar_wall"]["disk_roles"]`` doesn't exist for my node yet, that means user didn't set it and I don't need to add it into this hash

tboerger: You will do it anyway somewhere. But feel free to take my 2 cents however you want :)

vuntz:
A couple of comments on the UI:
- can we move this new bit just after "intended role" (and before availability zone)?
- do we really need to put a disabled text field for the disk names instead of just some text?
- would it be possible to put the non-unique disk name (sda) and the size of the disk somewhere, so that it helps people find the right disk?

crowbar_framework/app/views/nodes/_form.html.haml
```diff
@@ -113,6 +113,35 @@
...
+              = t(".disk_role")
+        %tbody
+          - (@node.crowbar_wall["claimed_disks"] || {}).each do |disk_name, disk|
+            %tr
+              %td
+                %span.form-control{ :disabled => "disabled" }
+                  = disk_name
+              %td
+                %span.form-control{ :disabled => "disabled" }
+                  = disk["owner"]
```
vuntz: By putting the string as is, then we don't use the nice name that you have put in the en.yml files. This will be very visible for DRBD ;-)

crowbar_framework/app/models/node_object.rb
```diff
@@ -256,6 +256,40 @@ def intended_role=(value)
...
+        crowbar_wall["disk_roles"].delete(name)
+      end
+    end
+  end
+
+  def add_unclaimed_disks_to_roles
+    # Check for current unclaimed disks, and add unclaimed disk to
+    # disk_roles hash if it is not there yet.
+    # (New disk could have been added to the node)
+    crowbar_wall["disk_roles"] ||= {}
```
vuntz: I would really do this at the beginning of the `disk_roles` method.

crowbar_framework/app/helpers/nodes_helper.rb
vuntz: I would agree with @tboerger; I understand that the goal is to avoid running code uselessly, but the way it reads is a bit convoluted. Maybe we should rethink the changes you did in node_object.rb and make the disk_roles method just return the attribute, and add a new `refresh_disk_roles` method?

crowbar_framework/app/views/nodes/_form.html.haml
jsuchome: oh, right

crowbar_framework/app/helpers/nodes_helper.rb
tboerger: That sounds pretty good

crowbar_framework/app/models/node_object.rb
tboerger: :+1:

jsuchome:
@vuntz:
> can we move this new bit just after "intended role" (and before availability zone)?

Done

> do we really need to put a disabled text field for the disk names instead of just some text?

Done
> would it be possible to put the non-unique disk name (sda) and the size of the disk somewhere, so that it helps people find the right disk?

Hm. Actually, not only ``disk_roles``, but also that ``claimed_disks`` map that I'm using is only working with unique names, so I would need to adapt both those maps, and I'm not sure if it is worth it

kbaikov:
UI note: Disk roles are visible only during node edit. But should be visible in view mode also. (i.e. before clicking edit node)

crowbar_framework/app/helpers/form_helper.rb
```diff
@@ -42,6 +42,31 @@ def roles_for_select(selected)
     )
   end
 
+  def beautify_disk_owner(owner)
+    owner2label = {
```
tboerger:
```
@owner_label ||= {
}
```

crowbar_framework/app/helpers/form_helper.rb
```diff
@@ -42,6 +42,31 @@ def roles_for_select(selected)
...
+      "Ceph" => t(".disk_role_ceph"),
+      "Cinder" => t(".disk_role_cinder"),
+      "Swift" => t(".disk_role_swift"),
+      "LVM_DRBD" => t(".disk_role_drbd"),
+      "sbd" => t(".disk_role_sbd")
+    }
+    owner2label[owner] || owner
```
tboerger: Maybe 

```
@owner_label.fetch(owner, owner)
```

jsuchome:
> UI note: Disk roles are visible only during node edit. But should be visible in view mode also. (i.e. before clicking edit node)

Good idea. But couldn't the list of disk be too long? (User could live with that in specialized edit view, but in show view...?)

tboerger:
I'm fine with that. We can improve it later on anyway. :+1:

rsalevsky:
For the node_object model it should be possible to write tests without much overhead.

crowbar_framework/app/views/nodes/_form.html.haml
```diff
@@ -105,6 +105,39 @@
...
+            %th.col-sm-3
+              = t(".disk_unique")
+            %th.col-sm-2
+              = t(".disk_role")
+        %tbody
+          - all_disks = all_disks_info
+          - (@node.crowbar_wall["claimed_disks"] || {}).each do |unique_name, disk|
+            %tr
+              %td
+                = all_disks.fetch(unique_name, unique_name)
```
vuntz: I wonder if the fallback should really be `unique_name`. Shouldn't we just put the usual long dash that is displayed when we don't have data? The issue is that we already display `unique_name` on the column after that.

vuntz:
Just to clarify: we don't display the disk roles on the node info page (not the edit one); is that intended?

jsuchome:
> Just to clarify: we don't display the disk roles on the node info page (not the edit one); is that intended?

That;s what Konstantin already mentioned as well. I'm not sure if it wouldn't get too crowded in case there are many disks. So I'd postpone this to some next PR. Maybe even consulting with designer would be good

crowbar_framework/app/views/nodes/_form.html.haml
jsuchome: That makes sense

vuntz:
> That;s what Konstantin already mentioned as well. I'm not sure if it wouldn't get too crowded in case there are many disks. So I'd postpone this to some next PR. Maybe even consulting with designer would be good

Works for me; make sure to create a follow-up card then :-)

rsalevsky:
@jsuchome, this repository got merged into [crowbar-core](https://github.com/crowbar/crowbar-core).

jsuchome:
Closing, as there's already PR to new repo

PR #1309, opened 2015-07-09
Remove extra ProposalMethods module

bkutil:
This module was used as a 'common' include during ProposalObject
and Proposal switch, to keep the API equivalent. It is no longer needed and the methods were moved
into the Proposal model.


tboerger:
Beside the style offenses :+1:

MaximilianMeister:
just moving code around, so style offenses were not really introduced
:+1: 
merging

PR #1311, opened 2015-07-10
Relax some rubocop rules like line length

tboerger:
Additional to that I relaxed the AbcSize as this is currently far too much refactoring.


rsalevsky:
Looks got to me, but we should wait with merging until next week.

tboerger:
I would take this for the meantime to stop spaming of HoundCI partially.

MaximilianMeister:
:+1:

PR #1312, opened 2015-07-10
UI for claiming disks from node editing

jsuchome:
UI for claiming disks from node editing page

Alternative approach to https://github.com/crowbar/barclamp-crowbar/pull/1306


tboerger:
It should be named disk_roles and not disks_roles

jsuchome:
@tboerger fine, I'm happy to rename it, but first we must decide if this or https://github.com/crowbar/barclamp-crowbar/pull/1306 should be used

vuntz:
As discussed last week, I feel the other approach is actually better.

PR #1315, opened 2015-07-14
Tighten up Crowbar::Product::is_ses? check

tserong:
Previously this would just check if suse_enterprise_storage was in the
BarclampCatalog.  Unfortunately, this will always be true if the Ceph
barclamp is installed, as it's a member of suse_enterprise_storage since:

  https://github.com/crowbar/barclamp-ceph/pull/154

See subsequent discussion at:

  https://github.com/crowbar/barclamp-crowbar/pull/1266

This fix tightents up the check to verify that suse_enterprise_storage
is a member of itself, which won't happen unless that barclamp is
really, truly installed.

Signed-off-by: Tim Serong <tserong@suse.com>

### DIFF
--- a/crowbar_framework/lib/crowbar/product.rb
+++ b/crowbar_framework/lib/crowbar/product.rb
@@ -17,7 +17,14 @@
 module Crowbar
   class Product
     def self.is_ses?
-      BarclampCatalog.barclamps.key?("suse_enterprise_storage")
+      # This checks if the product is SUSE Enterprise Storage.  It's not
+      # enough to just check if 'suse_enterprise_storage' is in BarclampCatalog,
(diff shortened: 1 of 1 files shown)


vuntz:
Assuming it works, +1 :-)

tserong:
I tested this manually using the rails console. After the above change, I get the following:

* neither ceph barclamp nor ses barclamp installed: is_ses? returns false
* ceph barclamp installed, ses barclamp not installed: is_ses? returns false (this is my cheapass simulation of a cloud deploy)
* ceph barlcamp and ses barclamp both installed: is_ses? returns true

toabctl:
+1 (but haven't tested it)

PR #1316, opened 2015-07-19
Allow ohai 6.24.x

dirkmueller:
Also sync requirements between Gemfile and config/boot.rb

### DIFF
--- a/crowbar_framework/Gemfile
+++ b/crowbar_framework/Gemfile
@@ -35,7 +35,7 @@ gem "simple_navigation_renderers", "~> 1.0.2"
 gem "sqlite3", "~> 1.3.9"
 gem "syslogger", "~> 1.6.0"
 
-gem "ohai", "~> 6.22.0"
+gem "ohai", "~> 6.24.2"
 gem "chef", "~> 10.32.2"
 
 gem "mixlib-shellout", "~> 1.4.0",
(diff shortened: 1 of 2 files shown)


tboerger:
:-1: It's ok to update gem versions but let's keep the simple versioning in boot.rb because this get's used for our packages only, so if the matching package have been installed.

dirkmueller:
@tboerger it doesn#t look like the versioning gets used at all in config/boot.rb, since it was wrong and nothign seems to have complained about it. how about we simply remove it to not duplicate versioning?

tboerger:
It IS used. It's used in the case of packaging only (as we are not installing Gemfile, it's only installed with crowbar-barclamp-crowbar-devel). So these part of boot.rb makes definitly sense. It's not just a joke from my side ;)

tboerger:
The versioning of boot.rb should stay as it can use totally wrong versions if somebody installs a more current version. Even if it's just by accident.

rsalevsky:
Please rebase

tboerger:
Still :-1: as stated above.

dirkmueller:
@tboerger as seen in https://github.com/crowbar/barclamp-crowbar/pull/1319 the enforcement was not effective before.

tboerger:
> @tboerger as seen in #1319 the enforcement was not effective before.

It have been effective, but this was just a typo as the ```version:``` key was missing.

rsalevsky:
+1

tboerger:
:+1:

tboerger:
Even if the boot.rb change is not a hard requirement.

PR #1321, opened 2015-07-21
Do not try to stop/start chef-client daemon on windows nodes

dirkmueller:
For windows nodes, we rely on the regular chef-client run and we don't
connect through ssh.

http://bugzilla.suse.com/show_bug.cgi?id=937760
(cherry picked from commit 47d35423f547a5b3cdcbc4564dfca86b55a9988c)


bmwiedemann:
:+1: code looks good. did not test

MaximilianMeister:
@bmwiedemann It's a backport, so the code has already been reviewed.

+1

PR #1322, opened 2015-07-21
Switch to http for rubygems, update schema.rb

tboerger:


rsalevsky:
+1

crowbar_framework/Gemfile
```diff
@@ -15,7 +15,7 @@
 # limitations under the License.
 #
 
-source "https://rubygems.org"
+source "http://rubygems.org"
```
vuntz: If we're only doing it because SLE11 is broken, then I suggest to wait for the move to SLE12. Is there any other reason?

(it's bad to go back to http)

rsalevsky: It's just because SLE11 is broken. Due to the move is planned for the next sprint I'm fine with waiting.

tboerger: Switching to http is not a real issue and not too bad as we never send any sensitive informations. It grabs only some additional rubygems used for the development environment as you have dropped the development and test dependencies from the devel subpackage.

MaximilianMeister:
why is the update of the schema needed? maybe add sth to the commit message

tboerger:
> why is the update of the schema needed? maybe add sth to the commit message

It's just getting updated that way if you migrate the database. It's automatically done by rails.

crowbar_framework/Gemfile
vuntz: This is not about encrypting, it's about checking that the server is the one it claims to be.

tboerger: IMHO it's still not too bad for the short time we need this change as this anyway gets executed only on the admin node within our development environment used by 3 people.

dirkmueller: The 3 people should be using SLE12 .. :-) https://trello.com/c/2sw8Wc0M/426-5-p9-make-crowbar-rails-app-work-on-sle12

PR #1324, opened 2015-07-21
Sync Gemfile requirements with config/boot.rb

dirkmueller:
Also sync requirements between Gemfile and config/boot.rb

### DIFF
--- a/crowbar_framework/config/boot.rb
+++ b/crowbar_framework/config/boot.rb
@@ -27,74 +27,74 @@ if File.exists? ENV["BUNDLE_GEMFILE"]
   Bundler.require(:default, Rails.env)
 else
   # rails related
-  gem "rails", version: "~> 4.1"
+  gem "rails", version: "~> 4.1.9"
   require "rails/all"
 
-  gem "haml-rails", version: "~> 0.5"
+  gem "haml-rails", version: "~> 0.5.3"
   require "haml-rails"
 
-  gem "sass-rails", version: "~> 4.0"
+  gem "sass-rails", version: "~> 4.0.5"
   require "sass-rails"
 
-  gem "rainbows-rails", version: "~> 1.0"
+  gem "rainbows-rails", version: "~> 1.0.1"
   require "rainbows-rails"
 
   # general stuff
-  gem "activerecord-session_store", version: "~> 0.1"
+  gem "activerecord-session_store", version: "~> 0.1.0"
   require "activerecord/session_store"
 
-  gem "active_model_serializers", version: "~> 0.9"
+  gem "active_model_serializers", version: "~> 0.9.0"
   require "active_model_serializers"
 
-  gem "activeresource", version: "~> 4.0"
+  gem "activeresource", version: "~> 4.0.0"
   require "active_resource"
 
-  gem "closure-compiler", version: "~> 1.1"
+  gem "closure-compiler", version: "~> 1.1.10"
   require "closure-compiler"
 
-  gem "dotenv", version: "~> 1.0"
+  gem "dotenv", version: "~> 1.0.2"
   require "dotenv"
 
-  gem "hashie", version: "~> 2.1"
+  gem "hashie", version: "~> 3.3.1"
   require "hashie"
 
-  gem "i18n-js", version: "~> 2.1"
+  gem "i18n-js", version: "~> 2.1.2"
   require "i18n-js"
 
-  gem "js-routes", version: "~> 0.9"
+  gem "js-routes", version: "~> 0.9.7"
   require "js-routes"
 
-  gem "kwalify", version: "~> 0.7"
+  gem "kwalify", version: "~> 0.7.2"
   require "kwalify"
 
-  gem "mime-types", version: "~> 1.25"
+  gem "mime-types", version: "~> 1.25.1"
   require "mime/types"
 
-  gem "redcarpet", version: "~> 3.2"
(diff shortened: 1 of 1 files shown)


tboerger:
Ok, you are a new rails specialist? Still :-1: from my side.

dirkmueller:
@tboerger Did I say that I'm the rails specialist anywhere? 

@vuntz @rsalevsky @MaximilianMeister any other opinions ? It seems not possible to discuss the technical matter with @tboerger

rsalevsky:
I have too less background information to make a real decision. All what I know is that we have the versions in this file to avoid loading the wrong gem versions. It can happen that a customer installs accidentally another gem version via `gem install` or `bundler`. Specially this can also happen in the dev setup (where we have both setups production and development) for testing e.g. a newer gem version. When we adjust the gem version in `boot.rb` to the exact version then we need to ship a new crowbar update which each gem update otherwise it can happen that the version does not longer match and crowbar can not start.

vuntz:
> It can happen that a customer installs accidentally another gem version via gem install or bundler

Won't happen for a customer, but I guess it can happen for a developer.

But the question is why do it one way in Gemfile and another way in boot.rb? Wouldn't it be best to be consistent? There might be a good reason, but not being an expert in that area, I don't know it...

Also, does it really hurt to specify the exact version we want?

tboerger:
> But the question is why do it one way in Gemfile and another way in boot.rb? Wouldn't it be best to be consistent? There might be a good reason, but not being an expert in that area, I don't know it...

Initially I have done it the same for Gemfile, but this have been set more restrictive to force Travis for better matching gem versions.

> Also, does it really hurt to specify the exact version we want?

If you don't mind to release a new crowbar package just because some rubygems have been updated because of security issues or whatever feel freer to take it as it is. I have taken the lesser restrictions to keep updates simple. But in the case of this "hard" version constraints we can directly skip the boot.rb version definitions and ship the Gemfile within the normal package as the differences are gone.

Your choice...

aspiers:
:-1: simply because the commit message doesn't explain the rationale for the change.  This is always important, but *especially* when the change is controversial which this one seems to be :-)

Whilst not a Rails guru like @tboerger I do have quite a lot of experience with the Ruby ecosystem and [with gem packaging / versioning in particular](https://en.opensuse.org/index.php?title=openSUSE:Ruby_Gem_Strategies&action=history), and if you can help me understand the exact problem I will do my best to help us reach a consensus.

aspiers:
And BTW here is some required reading for anyone in this discussion ...

* https://en.opensuse.org/openSUSE:Ruby_Gem_Strategies
* http://ourobengr.com/2013/10/tied-to-the-rails/#more-632

dirkmueller:
@aspiers commit message updated

dirkmueller:
> If you don't mind to release a new crowbar package just because some rubygems have been updated
>  because of security issues or whatever feel freer to take it as it is.

First of all this change does not lock the rubygems to a specific version. What it does is it locks it to a specific minor version. So if we have  a security update in rubygem foo and that security fix can be applied by upgrading from version 1.2.4 to 1.2.5, we can do that without touching crowbar. 

In any case this reason is not very valid for me: QA is retesting crowbar *anyway* when there is a any package change in any rubygem that crowbar depends on. So whether they retest crowbar with or without this dependency change makes absolutely zero difference from the work required.

aspiers:
Yeah @dirkmueller that sounds reasonable to me.

The comments from @vuntz, @tboerger, and @rsalevsky referring to "exact versions" and "hard version constraints" make me wonder - you all realise that this is not changing `~>` to `==`, right?  It's merely adding one extra level of versioning.  Like Dirk said, `~> 4.1.9` will still allow 4.1.10, 4.1.11 etc.  So I'm wondering why you object to it.  `rails ~> 4.1.9` makes more sense to me than `rails ~> 4.1` since it prevents accidental minor version upgrades e.g. to Rails 4.2.x.  Even though technically with semver those upgrades should be safe, we all know they're often not in the real world.  And consistency with `Gemfile` seems a reasonable goal too.

I could well be missing something, but so far I didn't see a clear explanation of why this would be a bad change.

tboerger:
So just drop the boot.rb stuff und ship the damn Gemfile. I'm out.

aspiers:
@tboerger How would shipping the "damn Gemfile" help? Wouldn't that require starting to use bundler?

rsalevsky:
@aspiers No, a gemfile just descries dependencies. This does not include using bundler, when you look a the boot.rb then you see that we use the versions from the gemfile when available.

aspiers:
> This does not include using bundler

I'm pretty sure it does; see below:

> when you look a the boot.rb then you see that we use the versions from the gemfile when available.

Yes, but via bundler:

    https://github.com/crowbar/barclamp-crowbar/blob/master/crowbar_framework/config/boot.rb#L23

and if we use bundler, we have to decide [when to generate `Gemfile.lock`](https://en.opensuse.org/openSUSE:Ruby_Gem_Strategies#When_.2F_how_should_Gemfile.lock_be_generated.3F). 

So as you can see this is far from a simple decision.  I would recommend everyone reads that wiki page, and then we have a meeting to discuss.

rsalevsky:
@dirkmueller, this repository got merged into [crowbar-core](https://github.com/crowbar/crowbar-core).

dirkmueller:
superceded by https://github.com/crowbar/crowbar-core/pull/56

PR #1325, opened 2015-07-22
Use apache as an proxy for the rails app

tboerger:
That way we get rid of the useless "please click here to get to the administration server" page and get access to crowbar directly.


MaximilianMeister:
Assuming it works +1

but maybe add a "Closes #1174" to the commit message, to keep the relation to the previous discussion of the other approach

tboerger:
Updated commit message.

aspiers:
Why use Apache as a proxy when you could just make it do a 308 Permanent Redirect?  Doesn't that just add extra load to the admin server for no benefit?

rsalevsky:
We need this anyway when we switch to https and my experience says the load is negligible.

tboerger:
> Why use Apache as a proxy when you could just make it do a 308 Permanent Redirect? Doesn't that just add extra load to the admin server for no benefit?

Because in another step we can even enable SSL certs directly through apache, in a case of redirect we will lose this ability again. Just in the case we will stick with rainbows :).

we should consider to switch to puma like it is already the case for hawk :).

vuntz:
Please listen on port 3000 only on 127.0.0.1 then.

tboerger:
> Please listen on port 3000 only on 127.0.0.1 then.

That's a valid point and a good idea. should i do that within that pull request?

vuntz:
> That's a valid point and a good idea. should i do that within that pull request?

That'd be great, yes!

tboerger:
> That'd be great, yes!

Ha, but that will be difficult on the short run. All scripts currently rely on port 3000 :(

MaximilianMeister:
mkcloud/qa_crowbarsetup would need some minor adaptions as well, as there is some usage of port 3000

tboerger:
> mkcloud/qa_crowbarsetup would need some minor adaptions as well, as there is some usage of port 3000

That's no problem as these commands are executed locally on the admin node. Tools like crowbar_register or autoyast needs some love to be aware of that.

chef/cookbooks/crowbar/recipes/default.rb
```diff
@@ -289,7 +289,7 @@
   owner "crowbar"
   group "crowbar"
   mode "0644"
-  variables(:web_host => "0.0.0.0",
+  variables(:web_host => "127.0.0.1",
```
tboerger: I will ignore that as it gets autofixed with the merge.

tboerger:
@vuntz I think I have created all required pull requests and updated this one for listening only on 127.0.0.1 with the rails application.

dirkmueller:
Thanks for getting rid of the ugly placeholder page! 

+1 code wise, not tested though.

vuntz:
Please do not merge for now; we'd like to get something working for M1 first.

tboerger:
> Please do not merge for now; we'd like to get something working for M1 first.

Ok, so take care about it after M1 as I will be on vacation :)

PR #1327, opened 2015-07-22
Move BDD to features, add cucumber

tboerger:
This is my proposal for getting rid of the old erlang parts and be prepared for implementing integration tests on cucumber.


dirkmueller:
how does that work packaging wise? do we now depend on cucumber as well?

tboerger:
It's no runtime dependency, it should be a devel dependency like rspec.

MaximilianMeister:
LGTM

@tboerger so your thought is to put the crowbar related features here in this repo, and not in the cct repo? to me that would make sense.

@tboerger are those features that you moved in the 2nd commit API tests? they sound a bit like UI tests so they would need a WebDriver... if they are about UI, then do we want to integrate that in the future? or why did you keep them?

maybe @vmoravec can have a look here

tboerger:
@MaximilianMeister cct is more some kind of integration testing including openstack. Currently i have just moved them to a proper place and we get rid of the erlang code nobody wants to have.

It makes totally sense to write features for Crowbar only directly within crowbar instead of relying on another tool. That way we can run these features directly on Travis and get even better feedback from Travis. As features are in general UI integration tests we need some driver on the long term, the easiest solution should be the usage of PhantomJS which can be integrated quite easy. But this is a different thing we need to consider when we start to implement the step definitions.

dirkmueller:
+1

vmoravec:
Can't comment on the erlang code, but for the rest +1

PR #1330, opened 2015-07-23
Log errors backtraces

bkutil:
Errors should not be silenced in service object and backtrace
should be always logged.


tboerger:
:+1:

rsalevsky:
+1

aspiers:
It would have been nicer to avoid the duplicated code, e.g. by extending `Rails.logger` with a new `#error_with_backtrace` method.

PR #1331, opened 2015-07-23
Environment config updates

tboerger:
* Raised crowbar version
* Added initial rainbows worker

### DIFF
--- a/crowbar_framework/config/crowbar.env
+++ b/crowbar_framework/config/crowbar.env
@@ -1,3 +1,5 @@
+export RAINBOWS_WORKERS=5
+
 export CROWBAR_VERSION=1.10
 export CROWBAR_LOG_DIR=/var/log/crowbar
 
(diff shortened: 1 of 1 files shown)


crowbar_framework/config/crowbar.env
```diff
@@ -1,4 +1,6 @@
-export CROWBAR_VERSION=1.10
+export RAINBOWS_WORKERS=5
+
+export CROWBAR_VERSION=3.0
```
dirkmueller: maybe 2.95 until we released 3.0 ?

tboerger: But we are working on 3.0, this is the upcoming version :)

rsalevsky:
+1

crowbar_framework/config/crowbar.env
```diff
@@ -1,4 +1,6 @@
-export CROWBAR_VERSION=1.10
+export RAINBOWS_WORKERS=5
```
MaximilianMeister: @tboerger does the `RAINBOWS_WORKERS` env influence only development or also production?

tboerger: Just development as production uses a custom rainbows config.

MaximilianMeister:
> Just development as production uses a custom rainbows config.

okay thanks.

:+1: then

PR #1333, opened 2015-07-24
Use default_os attribute from provisioner as default platform

vuntz:
It turns out that there is an attribute for defining the default
platform, and therefore we should not just rely on the OS of the admin
server.

This also means that the variable cannot be cached forever since the
attribute might change.


tboerger:
Some small improvement, after that :+1:

vuntz:
@tboerger done

vuntz:
Tested, and it works as intended.

rsalevsky:
+1

dirkmueller:
LGTM

bmwiedemann:
:+1:

PR #1334, opened 2015-07-28
eventually use CROWBAR_PORT from the environment

MaximilianMeister:
in the development environment you may want to use a different port

### DIFF
--- a/bin/barclamp_lib.rb
+++ b/bin/barclamp_lib.rb
@@ -25,7 +25,7 @@ require 'getoptlong'
 @debug = false
 @hostname = ENV["CROWBAR_IP"]
 @hostname = "127.0.0.1" unless @hostname
-@port = 3000
+@port = ENV["CROWBAR_PORT"] || 3000
 @headers = {
   "Accept" => "application/json",
   "Content-Type" => "application/json"


rsalevsky:
-1 We have already the `--port` option.

MaximilianMeister:
@rsalevsky and why shouldn't you be able to set it via an the environment? and why is it set like this in `crowbar_machines` as well?

rsalevsky:
Having both makes no sense and cause confusion, maybe also conflicts with other variables due to they are global. In general we should use less ENV variables as possible.

`crowbar_machines` is a other topic the new client will use correct parameters

aspiers:
> Having both makes no sense and cause confusion, maybe also conflicts with other variables due to they are global. In general we should use less ENV variables as possible.

In general I totally disagree - supporting both CLI options and environment variables is a very common pattern.  git and the openstack CLI are just two of many examples of very well-known software which does this.  It makes plenty of sense, especially to avoid repeating having to specify the same CLI options over and over again.  The concern about name conflicts does not apply due to the `CROWBAR_` prefix.

It's important to note that the CLI option should always take precedence over the corresponding environment variable - again this is conventional and makes more sense since it allows setting the defaults in the environment and then overriding on the CLI only when necessary.

tboerger:
This should be obsolete with crowbar-client but :+1: as it doesnt hurt.

MaximilianMeister:
rebased to the port 80 change

MaximilianMeister:
okay...  **un**rebased to the port 80 change...

vuntz:
+1

PR #1278, opened 2015-05-28
document YAML requirement to quote alias strings

aspiers:
The @ and ` characters can be used anywhere in a plain string except at the beginning, because they are reserved indicators:

  http://yaml.org/spec/1.2/spec.html#id2774228

See https://bugzilla.suse.com/show_bug.cgi?id=932619 for the corresponding change in product documentation.

PR #1335, opened 2015-07-29
batch export: don't strip double-quotes from aliases

aspiers:
`@` is a reserved indicator in YAML and can't start a plain scalar:

  http://yaml.org/spec/1.2/spec.html#id2774228

Therefore aliases need to be wrapped in double-quotes.  This was already fixed in the documentation for `crowbar batch build`:

  https://github.com/crowbar/barclamp-crowbar/pull/1278

but it appears that I deliberately made `batch export` strip the quotes in the original implementation (presumably on grounds of ugliness) before realising that that violated the YAML spec.

### DIFF
--- a/bin/crowbar_batch
+++ b/bin/crowbar_batch
@@ -44,7 +44,6 @@ require "easy_diff"
 require "chef/mixin/deep_merge"
 require "pp"
 
-ALIAS_REGEXP = /"(@@[^ @]+@@)"/
 ALIAS_TEMPLATE = '@@%s@@'
 
 INDENT = '   '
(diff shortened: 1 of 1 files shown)


thutterer:
+1
I've tested this as a patch just now and it works. Without the patch, the exported YAML was broken because of list entries like `- @@dashboard@@`.
With the patch those entries now get quoted correctly: `- "@@dashboard@@"`.

rsalevsky:
Tested the patch with the same result. Thanks! :+1:

dirkmueller:
+1

PR #1336, opened 2015-07-30
prevent crowbar batch from locking itself and the user out

aspiers:
If crowbar batch is used to export the crowbar proposal from one cloud and apply it to another, it will change the machine-install user's password, which is typically the one used to run crowbar batch.  This would break any future HTTP digest authentications against the Crowbar REST API, including ones invoked immediately after the change, by the still running crowbar batch process.  So if the user inadvertently attempts to change the password they're currently relying on, simply output a warning and ignore the attempt.

Also improve error reporting when retrieving aliases.

### DIFF
--- a/bin/crowbar_batch
+++ b/bin/crowbar_batch
@@ -313,20 +313,38 @@ def wipe_attribute(new_json, barclamp, attribute)
 end
 
 def merge_attributes(new_json, barclamp, proposal)
+  attrs = proposal["attributes"]
+
   to_merge = {
     "attributes" => {
-      barclamp => proposal["attributes"]
+      barclamp => attrs
     },
     "deployment" => {
       barclamp => proposal["deployment"]
     },
   }
 
+  prevent_password_lockout(attrs) if barclamp == 'crowbar'
+
   # easy_merge! seems to have problems with Arrays of Hashes :-/
   #new_json.easy_merge! to_merge
 
   #new_json.extend Chef::Mixin::DeepMerge
   Chef::Mixin::DeepMerge.deep_merge!(to_merge, new_json)
+end
+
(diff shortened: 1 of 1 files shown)


aspiers:
This is not right yet.

aspiers:
OK this should be right now - needs a quck test though.

bin/crowbar_batch
```diff
@@ -329,6 +333,20 @@ def merge_attributes(new_json, barclamp, proposal)
...
+
+  users = attrs['users']
+  return unless users && users[@username]
+
+  if @password != users[@username]['password']
+    warn "Refusing to change password for '#{@username}' user " + \
+         #"from #{@password} to #{users[@username]['password']} " + \
+         "which would lock myself out!"
+  end
+  to_merge['attributes'][barclamp]['users'].delete @username
```
thutterer: While testing this I got a 
```
/opt/dell/bin/barclamp_lib.rb:536:in `eval': undefined local variable or method `to_merge' for main:Object (NameError)
```
Changing it to an instance variable `@to_merge` (in all its lines ofc) fixes this.

Second issue in this line: `barclamp` is unknown as-well. Making it a method argument fixed it for me.

aspiers: Thanks for the catch but there is a better fix for than these ;-)

thutterer:
Tested again. Works.
+1

rsalevsky:
LGTM but I have not much experience with crowbar_batch

rsalevsky:
@aspiers, this repository got merged into [crowbar-core](https://github.com/crowbar/crowbar-core).

aspiers:
Superseded by https://github.com/crowbar/crowbar-core/pull/35

PR #1337, opened 2015-07-30
batch build: fix merging of Array attributes

aspiers:
I can't remember why I ever avoided `#easy_merge!` (**UPDATE:** I remembered), but it's now clear that combining `#easy_diff` and `#deep_merge!` was a bad idea, as explained in the new comments.

**N.B. requires https://github.com/Blargel/easy_diff/pull/5 in order to work properly, so shouldn't be merged until that's merged or we've at least patched our easy_diff package.**

Part of https://trello.com/c/ssI9nw9p

### DIFF
--- a/bin/crowbar_batch
+++ b/bin/crowbar_batch
@@ -41,7 +41,6 @@ require "open3"
 require "tempfile"
 
 require "easy_diff"
-require "chef/mixin/deep_merge"
 require "pp"
 
 ALIAS_TEMPLATE = '@@%s@@'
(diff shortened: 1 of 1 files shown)


rsalevsky:
+1

aspiers:
@thutterer found the case where `#easy_merge` fails with Arrays of Hashes.  I am fixing the bug in the `easy_diff` gem, but this is not safe to merge yet.

aspiers:
OK, this requires https://github.com/Blargel/easy_diff/pull/5 in order to work properly.

aspiers:
This can be safely merged when https://build.suse.de/request/show/64242 is accepted.

aspiers:
The SR has been accepted so this is safe to merge.  Just need one more +1 - preferably @vuntz since he's maybe the only other person who understands how `crowbar_batch` works ;-)

aspiers:
BTW presumably the gate failure is unrelated to this PR, since the gate doesn't yet use `crowbar_batch`.  I'm working on changing that though ;-)

thutterer:
Well, I'm not vuntz but I've tested this yesterday after manually updating the gem to 0.0.5 and applying this PR as a patch and it worked fine for me.
+1 (if this counts)

aspiers:
@thutterer thanks. What about your cinder validation issue - does that only happen with this patch applied?

thutterer:
Good question ... iirc I didn't get that far without your patch, but I'll double-check that and take back the +1 until I have an answer.

thutterer:
Wow @aspiers, it really is the case, that crowbar_batch is able to finish the cinder step when I'm NOT using the code from this pull request!
It still fails one step later with the old neutron "gre vs vlan" error that led to this pull request, of course.
So unfortunately I have to change my vote to -1.

aspiers:
I'm not as surprised as you are - they are fundamentally different mechanisms for merging attributes.  So you should run with `-d` in both cases and find out why the `easy_merge` version results in a broken set of attributes which fails the cinder validation tests.

aspiers:
(Or instead of `-d` you can simply add debugging to the relevant parts of crowbar batch's attribute merging code.)

tboerger:
:+1:

rsalevsky:
@aspiers, this repository got merged into [crowbar-core](https://github.com/crowbar/crowbar-core).

aspiers:
Superseded by https://github.com/crowbar/crowbar-core/pull/37

aspiers:
Superseded by crowbar/crowbar-core#39.

PR #1339, opened 2015-08-03
crowbar_batch build: improve YAML file error handling

thutterer:
- abort with a better error message if the exported YAML file contains aliases that are unknown to Crowbar
- abort if the YAML file is empy (bugfix)
- replace a warn&exit two-liner with a simple abort

### DIFF
--- a/bin/crowbar_batch
+++ b/bin/crowbar_batch
@@ -102,6 +102,7 @@ def build(input_file)
   end
 
   data = YAML.load(input_data)
+  abort "#{input_file} is empty" unless data
 
   global_options = data["global_options"] || {}
 
--- a/bin/crowbar_batch
+++ b/bin/crowbar_batch
@@ -96,6 +96,11 @@ def build(input_file)
     input_data.gsub!(ALIAS_TEMPLATE % aliaz, hostname)
   end
 
+  unknown_aliases = input_data.to_enum(:scan, ALIAS_REGEXP).map { Regexp.last_match.to_s }
+  unless unknown_aliases.empty?
+    abort "Some of the aliases in the YAML file are not known to Crowbar:\n- #{unknown_aliases.uniq.join("\n- ")}"
+  end
+
   data = YAML.load(input_data)
 
   global_options = data["global_options"] || {}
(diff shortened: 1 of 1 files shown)


aspiers:
+1

thutterer:
- abort if the file does not exist

aspiers:
Still +1

rsalevsky:
Codewise +1

bin/crowbar_batch
```diff
@@ -89,21 +89,27 @@ def build(input_file)
 
   host_by_alias = get_aliases[0]
 
-  input_data = File.open(input_file).readlines.join('')
+  abort "#{input_file} does not exist" unless File.file?(input_file)
```
tboerger: IMHO it's more conventional to use ```File.exist?```

bin/crowbar_batch
```diff
@@ -89,21 +89,27 @@ def build(input_file)
...
+  input_data = IO.read(input_file)
 
   # Translate @@alias@@ to Chef node name
   host_by_alias.each do |aliaz, hostname|
```
tboerger: I think you can write ```alias``` as well, maybe it's just highlighted bad in your editor.

bin/crowbar_batch
thutterer: ```File.exist?``` would also be true with directories.

bin/crowbar_batch
aspiers: `alias` is a Ruby keyword ... http://ruby-doc.org/core-2.2.2/doc/keywords_rdoc.html

tboerger: Really? Never heard about it :)

aspiers: Yeah, people write `aliaz` and `klass` for the same reason ...

thutterer:
@tboerger any objections to merge this as it is? it got +2
or do we want to wait for other people's review, @aspiers?

aspiers:
I already gave +1 ... twice :-)

thutterer:
> I already gave +1 ... twice :-)

I asked for your opinion to keep it open or merge it, not a third +1 :)

PR #1340, opened 2015-08-04
ExecStartPre requires an absolute path

MaximilianMeister:
http://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStart=

### DIFF
--- a/chef/cookbooks/crowbar/files/default/crowbar.service
+++ b/chef/cookbooks/crowbar/files/default/crowbar.service
@@ -7,7 +7,7 @@ Type=simple
 User=crowbar
 Group=crowbar
 PIDFile=/var/run/crowbar/crowbar.pid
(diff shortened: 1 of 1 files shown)


vuntz:
+1

chef/cookbooks/crowbar/files/default/crowbar.service
```diff
@@ -7,7 +7,7 @@ Type=simple
 User=crowbar
 Group=crowbar
 PIDFile=/var/run/crowbar/crowbar.pid
-ExecStartPre=rm -f /var/run/crowbar/looper-chef-client.lock /var/run/crowbar/chef-client.lock
+ExecStartPre=/bin/rm -f /var/run/crowbar/looper-chef-client.lock /var/run/crowbar/chef-client.lock
```
vuntz: Actually, should be /usr/bin/rm :-)

ovisan:
:+1:

PR #1344, opened 2015-08-10
replacing the queue databag with a rails activerecord model

MaximilianMeister:
I decided to put the properties into a text column which will then be serialized again, as we did it for the proposal model.
Once we have all data_bags and objects (node/role...) out of chef server we can design a proper model with all its relations between tables and columns.


crowbar_framework/app/models/proposal_queue.rb
```diff
@@ -0,0 +1,35 @@
...
+  self.primary_key = "name"
+
+  serialize :properties, Utils::JSONWithIndifferentAccess
+
+  def proposals
+    ProposalQueue.find("queue").properties["proposal_queue"]
+  end
+
+  def <<(item)
+    pq = ProposalQueue.find("queue")
```
aspiers: I'm not 100% sure but wouldn't `self.find` be better here? and same for below.

MaximilianMeister: makes sense, we can definitely do that.

crowbar_framework/app/models/proposal_queue.rb
```diff
@@ -0,0 +1,34 @@
...
+  class_attribute :chef_type
+
+  self.chef_type = "data_bag_item"
+  # only accept one default record named "queue"
+  self.primary_key = "name"
+
+  serialize :properties, Utils::JSONWithIndifferentAccess
+
+  def proposals
+    self.find("queue").properties["proposal_queue"]
```
aspiers: *sigh*  False positive from the dog...

crowbar_framework/app/models/proposal_queue.rb
MaximilianMeister: @aspiers should have tested it before. that actually doesn't work...
`NoMethodError: undefined method `find' for #<ProposalQueue:0x0000000559b7f8>`

aspiers: Sorry, I meant `self.class.find` ...

MaximilianMeister: then I would prefer it like it is. any other opinions?

MaximilianMeister: @aspiers we could also use a private method to just return an ActiveRecord object to get rid of the class identifier dups.

```ruby
...
private
  def proposal_queue
    ProposalQueue.find("queue")
  end
```

but that is a matter of taste I guess. anyone who has an opinion please share.

aspiers: Yeah, I'm really not sure which is best :)  Maybe we should ask @cornelius :)

cornelius: I would prefer using `ProposalQueue` over any of the `self` constructs, because it is more explicit. Repetition of the class name I wouldn't consider as an issue, because somehow you need to identify which class to call.

I would be concerned about the repetition of the `find` argument, because this looks like a potential source of errors, e.g. when mistyping "queue". So wrapping this in a method as proposed sounds like the right thing to me.

aspiers: Thanks @cornelius! I'm convinced :-)

tboerger:
Sorry but I really dislike this implementation. I'm fine with keeping the schema totally simple like it is for the beginning, but instead of creating a single mega record we should add a record for each queue item and iterate through these records ordered by id.

crowbar_framework/app/controllers/deploy_queue_controller.rb
```diff
@@ -25,6 +25,6 @@ def currently_deployed
   end
 
   def deployment_queue
-    Chef::DataBag.load("crowbar/queue")["proposal_queue"] rescue []
+    ProposalQueue.find("queue").proposals rescue []
```
rsalevsky: I prefer to handle this case in the proposals method, otherwise we tend to a fail driven code.

crowbar_framework/app/models/proposal_queue.rb
```diff
@@ -0,0 +1,40 @@
...
+  # only accept one default record named "queue"
+  self.primary_key = "name"
+
+  serialize :properties, Utils::JSONWithIndifferentAccess
+
+  def proposals
+    proposal_queue.properties["proposal_queue"]
+  end
+
+  def <<(item)
```
rsalevsky: Maybe we should replace this with a method named `push`. Then we use the same way as ruby handles arrays, when we need backward we can just set a alias with a depreciation warning.

tboerger: Even better should be an alias, but anyway I suggested to stop this behavior.

crowbar_framework/spec/models/proposal_queue_spec.rb
```diff
@@ -0,0 +1,55 @@
+require "spec_helper"
+
+describe ProposalQueue do
+  queue = JSON.parse(File.read("spec/fixtures/proposal_queue.json"))
+
+  let(:proposal_queue) { ProposalQueue.new(name: "queue", properties: queue) }
+  let(:proposal_queue_empty) do
+    ProposalQueue.new(name: "queue", properties: { "proposal_queue" => [] })
+  end
+  let(:proposal_queue_internals) { proposal_queue.properties["proposal_queue"] }
```
rsalevsky: This lines should be in a `before`.

crowbar_framework/spec/models/proposal_queue_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  context "proposal is queued" do
```
rsalevsky: You have no context so you can delete it.

crowbar_framework/lib/crowbar/deployment_queue.rb
```diff
@@ -19,9 +19,9 @@ class DeploymentQueue
 
     attr_reader :logger, :proposal_queue
 
-    def initialize(logger: Rails.logger, queue: ProposalQueue.new)
+    def initialize(logger: Rails.logger)
       @logger         = logger
-      @proposal_queue = queue
+      @proposal_queue = ::ProposalQueue.first
```
tboerger: Needs ordering by ```created_at```

crowbar_framework/lib/crowbar/deployment_queue.rb
```diff
@@ -130,29 +129,31 @@ def process_queue
...
           end
 
-          logger.debug("process queue: queue: #{proposal_queue.proposals.inspect}")
+          logger.debug("process queue: queue: #{@proposal_queue.inspect}")
 
           # Test for ready
           remove_list = []
-          proposal_queue.proposals.each do |item|
-            prop = Proposal.where(barclamp: item["barclamp"], name: item["inst"]).first
+          ProposalQueue.all.each do |item|
```
tboerger: Needs ordering by ```created_at```

crowbar_framework/lib/crowbar/deployment_queue.rb
MaximilianMeister: the order is always correct if we use the first record, as this is always the oldest

crowbar_framework/lib/crowbar/deployment_queue.rb
MaximilianMeister: same here `.all` retains the order FIFO just fine

crowbar_framework/lib/crowbar/deployment_queue.rb
tboerger: But that is never guaranteed, so please add it directly here or create a named scope ```ordered``` that you can call for get the FIFO record for sure, e.g.

```
  scope :ordered, lambda {
    order("proposal_queue.created_at ASC")
  }
```

crowbar_framework/lib/crowbar/deployment_queue.rb
tboerger: But that is never guaranteed, so please add it directly here or create a named scope ```ordered``` that you can call for get the FIFO record for sure, e.g.

```ruby
  scope :ordered, lambda {
    order("proposal_queue.created_at ASC")
  }
```

tboerger:
:+1:

rsalevsky:
:+1:

PR #1345, opened 2015-08-11
add yaml_db to the list of gems

MaximilianMeister:
* this is primarily needed to provide a rake task (db:data:dump)
  for the supportutils-plugin-suse-openstack-cloud but is also useful
  in general, as you can dump the crowbar database without having to
  specify the type of database (SQLite, MySQL, PostgreSQL etc...)

### DIFF
--- a/chef/cookbooks/crowbar/recipes/default.rb
+++ b/chef/cookbooks/crowbar/recipes/default.rb
@@ -78,6 +78,7 @@ when "suse"
     ruby2.1-rubygem-simple_navigation_renderers
     ruby2.1-rubygem-sqlite3
     ruby2.1-rubygem-syslogger
+    ruby2.1-rubygem-yaml_db
   )
   unless search(:node, "platform:windows").empty?
     pkglist.push "samba-client"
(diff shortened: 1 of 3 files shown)


MaximilianMeister:
this was the reason to include it
https://github.com/SUSE-Cloud/supportutils-plugin-suse-openstack-cloud/pull/42

MaximilianMeister:
is there still some cookbook/recipe where i need to install the package?

rsalevsky:
You need to add it to the packages and media.

MaximilianMeister:
@rsalevsky yes i know, this is done in parallel, but that was not my question. it was about cookbooks and package resources in the recipes

aspiers:
No, it should be a dependency of the crowbar package (not just the supportconfig plugin package since I guess other things could also use this rake task, e.g. the backup script).  Then it gets automatically installed when the admin node is installed.

aspiers:
+1

tboerger:
:+1:

dirkmueller:
+1

PR #1346, opened 2015-08-12
create a lockfile to prevent chef-daemon kick in during apply_role

MaximilianMeister:
the chef client will use this information to run only if the lock file is not present


crowbar_framework/lib/crowbar/lock_file.rb
```diff
@@ -0,0 +1,27 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+module Crowbar
+  class LockFile
+    def self.create(name)
+      File.new("tmp/#{name}.lock", "w")
```
rsalevsky: Do you really mean `tmp/` instead of `/tmp/`?

crowbar_framework/lib/crowbar/lock_file.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    def self.destroy(name)
+      File.delete("tmp/#{name}.lock")
```
rsalevsky: The check if the file exists is missing.

rsalevsky:
You should add some tests for the 'LockFile' class maybe with the help of https://rubygems.org/gems/fakefs.

aspiers:
@rsalevsky Yes that sounds like a good idea.

crowbar_framework/lib/crowbar/lock_file.rb
dirkmueller: predictable name in a world writeable directory is a no-go.

dirkmueller:
@MaximilianMeister I'm not sure what you're trying to achieve here.. creating a lock on teh admin node isn't needed..

crowbar_framework/lib/crowbar/lock_file.rb
MaximilianMeister: it is the tmp dir in the rails root

MaximilianMeister:
> I'm not sure what you're trying to achieve here.. creating a lock on teh admin node isn't needed..

@dirkmueller can you be a bit more specific what is needed then?

aspiers:
Yes @dirkmueller is right, the lock is on each Crowbar node!  Sorry for missing this.  If you don't know why then I strongly suggest re-reading the bug until you understand.  And BTW the commit messages are missing `bsc#` references and URLs.

crowbar_framework/lib/crowbar/lock_file.rb
aspiers: Rails root is the wrong machine as already mentioned.

aspiers:
This is not the right approach.

aspiers:
Need to change the locking mechanism to use `ssh_cmd` to write the lock on the remote node.

crowbar_framework/app/models/service_object.rb
```diff
@@ -822,9 +822,9 @@ def apply_role(role, inst, in_queue)
     # We also check that all nodes we'll require are in the ready state.
     #
 
-    # Initialize variables used in ensure at the end of the method
-    chef_daemon_nodes = []
-
+    # get all nodes except the admin node
+    nodes_without_admin = NodeObject.find_all_nodes.reject { |n| n.admin? }.map(&:name)
```
aspiers: I agree with the dog ;-)

aspiers:
Now looks very promising except the 2 nitpicks, thanks!

aspiers:
Uh-oh ... what about Windows nodes?  Do we have to worry about them?

MaximilianMeister:
still need to add some logging and an option to use remote or local lockfile creation. i would maybe do this with an inner class like this
``` ruby
module Crowbar
  class Lock
    class Remote
    end

    class Local
    end
  end
```

crowbar_framework/lib/crowbar/lock.rb
```diff
@@ -0,0 +1,55 @@
...
+        File.new("/var/lock/#{name}", "w")
+      end
+
+      def destroy(name)
+        @logger.info("Destroying local temporary lockfile #{name}")
+        File.delete("/var/lock/#{name}")
+      end
+    end
+  end
+end
```
MaximilianMeister: not sure if this is the most elegant way with 2 inner classes, please ruby specialists have a look. 
and i still don't like too much the 2 initializers, because they are DRY. can we share the logger somehow among the 2 inner classes? what would be the most elegant way to do that?

tboerger:
I would prefer a single class. If you stay with 2 classes you have to put them into separate files.

tboerger:
```ruby
class Lockfile
  attr_accessor :logger
  attr_accessor :path
  attr_accessor :remote

  def initialize(options = {})
    @logger = options.fetch :logger, Rails.logger
    @path = options.fetch :path, "/var/lock"
    @remote = options.fetch :remote, false
  end

  def create(name)
    if remote
    else
    end
  end

  def remove(name)
    if remote
    else
    end
  end

  def with_lock
    create
    yield
  ensure
    remove
  end

  protected

  def remote(command)
    # do it
  end

  def local(command)
    # do it
  end
end
```

```ruby
Lockfile.new.create("foo")
# complex block
Lockfile.new.remove("foo")
```

```ruby
Lockfile.new.with_lock do
  # simple block
end
```

This is how i think about it.

MaximilianMeister:
@aspiers so the caller needs to send the path `/var/chef/cache` and the file name `daemon-pause-file.lock` ?

MaximilianMeister:
> Uh-oh ... what about Windows nodes? Do we have to worry about them?

right, thanks. we have to filter them out

aspiers:
> @aspiers so the caller needs to send the path /var/chef/cache and the file name daemon-pause-file.lock ?

Yes but this should be done when the object is instantiated, not when the lock is created/removed.

crowbar_framework/lib/crowbar/lock.rb
```diff
@@ -0,0 +1,76 @@
...
+    attr_accessor :path
+    attr_accessor :remote
+
+    def initialize(options = {})
+      @logger = options.fetch :logger, Rails.logger
+      @path = options.fetch :path, "/var/lock"
+      @remote = options.fetch :remote, false
+    end
+
+    def create(name, nodes = [])
```
aspiers: The name should be set at instantiation not create/remove time.  It doesn't make sense to be able to use the same Lock object to create multiple lock files.

aspiers: Actually the list of `nodes` should probably be an instance variable too, to eliminate the risk of a different list being used for `#destroy`.

crowbar_framework/lib/crowbar/lock.rb
```diff
@@ -0,0 +1,76 @@
...
+      if @remote
+        remote(name, nodes, action: "create")
+      else
+        local(name, action: "create")
+      end
+    end
+
+    def destroy(name, nodes = [])
```
aspiers: This should just get the name (and probably the node list) from its internal instance variables set during `initialize`.

crowbar_framework/lib/crowbar/lock.rb
```diff
@@ -0,0 +1,76 @@
...
+      @logger = options.fetch :logger, Rails.logger
+      @path = options.fetch :path, "/var/lock"
+      @remote = options.fetch :remote, false
+    end
+
+    def create(name, nodes = [])
+      if @remote
+        remote(name, nodes, action: "create")
+      else
+        local(name, action: "create")
```
aspiers: Still strong -1 for local support unless you can give a concrete reason we need it.  Honestly, it looks to me like a waste of effort.  The locking process isn't even exclusive, which would lead to serious errors if anyone relied on it.

crowbar_framework/lib/crowbar/lock.rb
```diff
@@ -0,0 +1,76 @@
...
+      else
+        cmd, msg = "rm -f", "Destroying"
+      end
+      nodes.each do |node|
+        @logger.info("#{msg} remote temporary lockfile #{name} on #{node}")
+        RemoteNode.ssh_cmd(node, "#{cmd} #{@path}/#{name}")
+      end
+    end
+
+    def local(name, options = {})
```
aspiers: Please ditch this.

crowbar_framework/lib/crowbar/lock.rb
tboerger: I thought about setting remote to the list of nodes. The name makes sense, so just define the file instead of path for the initializer.

crowbar_framework/lib/crowbar/lock.rb
```diff
@@ -0,0 +1,76 @@
...
+    attr_accessor :remote
+
+    def initialize(options = {})
+      @logger = options.fetch :logger, Rails.logger
+      @path = options.fetch :path, "/var/lock"
+      @remote = options.fetch :remote, false
+    end
+
+    def create(name, nodes = [])
+      if @remote
```
tboerger: You can call just remote, there is a getter for that ;)

crowbar_framework/lib/crowbar/lock.rb
```diff
@@ -0,0 +1,76 @@
...
+        remote(name, nodes, action: "create")
+      else
+        local(name, action: "create")
+      end
+    end
+
+    def destroy(name, nodes = [])
+      if @remote
```
tboerger: You can call just remote, there is a getter for that ;)

crowbar_framework/lib/crowbar/lock.rb
```diff
@@ -0,0 +1,76 @@
...
+    ensure
+      destroy(name, nodes)
+    end
+
+    protected
+
+    def remote(name, nodes, options = {})
+      action = options.fetch :action
+      if action == "create"
+        cmd, msg = "touch", "Creating"
```
aspiers: `touch` is not good enough; the file should contain info which uniquely identifies its creator.  Then `destroy` should check that it owns the file before removing.  For debugging purposes it would also be useful if the file contains human-readable info about why it was created (e.g. the barclamp / proposal name and revision number at least).

tboerger:
If you create a ```Pathname``` out of path it's even easier to handle with ```join``` and create/destroy the files.

tboerger:
And please just use the getters. You can skip so many ```@``` that way.

crowbar_framework/lib/crowbar/lock.rb
```diff
@@ -0,0 +1,76 @@
...
+      else
+        cmd, msg = "rm -f", "Destroying"
```
aspiers: This would blindly allow a different consumer to remove a lock it doesn't own.

crowbar_framework/lib/crowbar/lock.rb
tboerger: If that is really useful i would append a ```claim``` attribute to the initializer.

aspiers: Also creating files is not an atomic action, so one consumer could overwrite the lock of another.  In contrast `symlink(2)` and `mkdir(2)` are atomic.

crowbar_framework/lib/crowbar/lock.rb
aspiers: Yeah that's fair - no need to split the directory from the filename.

crowbar_framework/lib/crowbar/lock.rb
aspiers: @tboerger yes the list of `initialize` parameters will indeed need to be extended.  I suggest `owner` and `reason`.

crowbar_framework/lib/crowbar/lock.rb
tboerger: Why? It's useful to have a local locking as well, there is no locking mechanism within rails, and if we already have it somewhere it makes sense to replace that with this.

crowbar_framework/lib/crowbar/lock.rb
tboerger: I suggested to create that!

crowbar_framework/lib/crowbar/lock.rb
aspiers: [We already have local locking](https://github.com/crowbar/barclamp-crowbar/blob/master/crowbar_framework/lib/file_lock.rb), and it actually works, unlike this implementation which is completely and utterly broken with respect to concurrency.  In fact it would be better to reuse that implementation by making it available on the other nodes and then keeping this class as a wrapper which executes it remotely via `ssh_cmd`.  Alternatively `ssh_cmd` could run `flock(1)` on the remote node.  The other suggestion I already made is to use `symlink(2)` or `mkdir(2)` which are both atomic operations (unlike just writing a file).

crowbar_framework/lib/crowbar/lock.rb
aspiers: Yes I know, but sorry, like I said above, it's totally redundant and broken w.r.t. concurrency.

crowbar_framework/lib/crowbar/lock.rb
tboerger: You are right for the class, I totally forgot that. So it makes more sense to add the remote functionality to that instead of creating another one.

aspiers: Agree in principle; the only problem is that "class" is actually just a collection of 4 class methods (presumably written by someone who didn't understand the difference between classes and modules).  And I can't see how any of the code could be reused when implementing the remote case.

tboerger: So take this pull request as an option to start small refactoring, the number of places where it's used is pretty small:

```
crowbar/crowbar_framework/app/models/service_object.rb:146: FileLock.acquire(name, :logger => @logger)
crowbar/crowbar_framework/app/models/service_object.rb:150: FileLock.release(f, :logger => @logger)
crowbar/crowbar_framework/app/models/role_object.rb:289: role_lock = FileLock.acquire "role:#{@role.name}"
crowbar/crowbar_framework/app/models/role_object.rb:311: FileLock.release role_lock
crowbar/crowbar_framework/app/models/role_object.rb:319: role_lock = FileLock.acquire "role:#{@role.name}"
crowbar/crowbar_framework/app/models/role_object.rb:322: FileLock.release role_lock
crowbar/crowbar_framework/lib/file_lock.rb:18: class FileLock
crowbar/crowbar_framework/lib/crowbar/deployment_queue.rb:231: FileLock
```

aspiers: No, this is definitely not the right PR to do that work, which is outside the scope of this sprint.  We need this remote locking urgently in lieu of a change we had to revert because it was bringing Jenkins to its knees, and we've already estimated that work.  So let's avoid getting distracted, get a simple solution for the remote locking working now, and have a separate story for tidying up the local locking later.

And anyway like I already said, I don't think there's any trivial way to reuse code between local and remote.  We'd need to automatically copy the library onto the other nodes, and there's an even more fundamental barrier which is that `FileLock` provides exclusive locking, whereas for this use case we need shared locking.

tboerger: Ok, i am out and stop reviewing that as nearly all comments on that pull request are "rejected" by you. Have fun with different classes for handling lockfiles.

aspiers: Rejected for concrete reasons which I clearly explained, not because I enjoy disputing comments / changes like you seem to be implying.  Yes, different classes makes sense when there is no easy way to reuse code between them.  Max will be having the fun though, not me, since I'm off on FTO now :)

rsalevsky:
@MaximilianMeister, this repository got merged into [crowbar-core](https://github.com/crowbar/crowbar-core).

MaximilianMeister:
Closing in favor of https://github.com/crowbar/crowbar-core/pull/5

PR #1348, opened 2015-08-13
Revert "apply_role: disable 15-minute chef run during apply_role"

dirkmueller:
Currently this causes a massive slowdown of proposal apply
up to a point where a HA deployment is no longer possible.
Revert until a proper fix is being found.

This reverts commit 41b6bca7f9e260a398df536015f22cca4c6c3f26.

### DIFF
--- a/crowbar_framework/app/models/service_object.rb
+++ b/crowbar_framework/app/models/service_object.rb
@@ -1109,10 +1109,6 @@ class ServiceObject
     end
     new_elements = expanded_new_elements
 
-    # stop chef daemon on all nodes
-    chef_daemon_nodes = new_elements.values.flatten
-    chef_daemon(:stop, chef_daemon_nodes)
-
     # save list of expanded elements, as this is needed when we look at the
     # old role
     if new_elements != new_deployment["elements"]
@@ -1423,9 +1419,6 @@ class ServiceObject
     restore_to_ready(all_nodes)
     process_queue unless in_queue
     [200, {}]
-  ensure
-    # start chef daemon on all nodes
-    chef_daemon(:start, chef_daemon_nodes)
   end
 
   def apply_role_pre_chef_call(old_role, role, all_nodes)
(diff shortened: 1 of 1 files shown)


aspiers:
+1 as discussed.

aspiers:
Actually, sorry - please update the commit message to reference https://bugzilla.suse.com/show_bug.cgi?id=936302

aspiers:
Also the commit reference is wrong; it actually reverts 41b6bca7f9e260a398df536015f22cca4c6c3f26 from #1277.

crowbar_framework/app/models/service_object.rb
```diff
@@ -1551,26 +1544,6 @@ def run_remote_chef_client(node, command, logfile_name)
     }
   end
 
-  def chef_daemon(action, node_list)
```
MaximilianMeister: wouldn't it be enough to just take out the caller method as i did in https://github.com/MaximilianMeister/barclamp-crowbar/commit/7b937bf64e14ca3c6b4eda169440b3ad7c661e31 as part of https://github.com/crowbar/barclamp-crowbar/pull/1346? maybe we could need this method later on.

aspiers: @MaximilianMeister That approach leads to a gradual accumulation of unused dead code, so it's not a good idea.  If we need it later we can easily retrieve from git.  But anyway this change is on a stable branch so it's only going to see bugfixes, so we probably won't ever need it again here.

MaximilianMeister: @aspiers then we should also remove `run_service` in the `node_object` as this is not used and `.service_command` in chef_object as this is not used either.

MaximilianMeister: ... just for rference, i mean https://github.com/crowbar/barclamp-crowbar/commit/75fbade214b9c85d7a1cb78831ddffc419a51454 and https://github.com/crowbar/barclamp-crowbar/commit/a4b4397595be5dfd0383b278a79da46dbb3b946a

dirkmueller: are you sure that nothing else uses them? I didn't want to take the risk right now..

dirkmueller:
@aspiers updated

toabctl:
+1

crowbar_framework/app/models/service_object.rb
aspiers: @dirkmueller @MaximilianMeister introduced them specifically for this purpose (in the same commit, even).  I just checked and can't see anything else using them.  But it doesn't matter too much if a few unused methods are left around in old stable branches.  The important thing is not to do that in `master`.

aspiers:
Should merge this when the gate passes.

aspiers:
Also we need to urgently get this merged so that all the gate builds get a lot quicker and Jenkins can chew through the backlog.

toabctl:
imo we can merge it now to speedup builds on gate.

PR #1350, opened 2015-08-14
ensure we install apache2 not lighttpd

aspiers:
On both SLE11 and SLE12, the package name is `apache2`.  Usually installing `httpd` works because the `apache2` package provides the `httpd` virtual package; however if there is a `SLE12-HA-Pool` repository available then `lighttpd` could accidentally get installed instead.

Unfortunately this missed Milestone 2 and appears to break Horizon when HA is used.  Tested on c19 and it works.


dirkmueller:
+1

PR #1355, opened 2015-10-28
webui: Only show non-allocated admin IP address on discovery

vuntz:
When we do hardware-installing/hardware-installed, then the node already
has an allocated IP address.

Backport of https://github.com/crowbar/crowbar-core/pull/98


jsuchome:
+1

tboerger:
:+1:

PR #1357, opened 2015-11-23
Implement upgrade state for nodes

nkrinner:
Implement a state to mark nodes ready for upgrades. Also make that
state visible in the UI


tboerger:
The image `blue.png` gets replaced afterwards by @kwwii?

crowbar_framework/app/models/service_object.rb
```diff
@@ -257,20 +257,34 @@ def remove_pending_elements(bc, inst, elements)
...
         node = NodeObject.find_node_by_name(node_name)
-        next if node.nil?
-
-        # Nothing to delay so mark them applying.
-        node.crowbar['state'] = 'ready'
-        node.crowbar['state_owner'] = ""
-        node.save
+        next if node.nil? or node.crowbar['state'] == 'upgrade'
+          # Nothing to delay so mark them applying.
+          @logger.info("nkrinner not changing node state to ready, as it has the upgrade tag: #{node.crowbar['name']} #{node.crowbar['state']}")
```
tboerger: `nkrinner`? :D

crowbar_framework/app/models/service_object.rb
```diff
@@ -257,20 +257,34 @@ def remove_pending_elements(bc, inst, elements)
...
+          restore_node(node)
       end
     end
   end
 
+  def remove_upgrade_state(nodes)
+    nodes.each do |node_name|
+      node = NodeObject.find_node_by_name(node_name)
+      next if node.nil?
+        @logger.info("nkrinner changing node state to ready, removing the upgrade tag: #{node.crowbar['name']} #{node.crowbar['state']}")
```
tboerger: `nkrinner`? :D

nkrinner:
Oh well, I still need to remove some debug output... :-)

crowbar_framework/app/models/node_object.rb
```diff
@@ -467,6 +469,7 @@ def ready?
 
   def state
     return 'unknown' if (@node.nil? or @role.nil?)
+    return 'upgrade' if self.crowbar['state'] === 'upgrade'
```
vuntz: There should be no need for this change, as the return value will be correct.

crowbar_framework/app/models/service_object.rb
```diff
@@ -257,20 +257,31 @@ def remove_pending_elements(bc, inst, elements)
     end
   end
 
+  def restore_node(node)
```
vuntz: Rename to `restore_node_to_ready`

crowbar_framework/app/models/service_object.rb
```diff
@@ -257,20 +257,31 @@ def remove_pending_elements(bc, inst, elements)
...
     with_lock "BA-LOCK" do
       nodes.each do |node_name|
         node = NodeObject.find_node_by_name(node_name)
-        next if node.nil?
-
-        # Nothing to delay so mark them applying.
-        node.crowbar['state'] = 'ready'
-        node.crowbar['state_owner'] = ""
-        node.save
+        next if node.nil? or node.crowbar['state'] == 'upgrade'
```
vuntz: s/or/||/

vuntz: Btw, why do we check for the upgrade state here? This method should only be called from `apply_role`, so it should not be relevant.

crowbar_framework/app/models/service_object.rb
```diff
@@ -257,20 +257,31 @@ def remove_pending_elements(bc, inst, elements)
...
+          # Nothing to delay so mark them applying.
+          restore_node(node)
       end
     end
   end
 
+  def remove_upgrade_state(nodes)
```
vuntz, line 277: Not sure where this will be called from, but sounds to me like you need to only do this for nodes which are in upgrade state.

crowbar_framework/app/models/service_object.rb
vuntz: Oh, it's because you set the state here: https://github.com/crowbar/barclamp-updater/pull/11/files#diff-cefc874d9c651cd9f09e3e298163248eR62

That's probably wrong. Will think about it.

crowbar_framework/app/models/service_object.rb
nkrinner, line 277: This is never called so far. i thought about adding a button in the node view to reset the node state.  Resetting the upgrade state will remove a few more steps too, as they don't have the chef-client running.

vuntz:
Oh, before I forget: one flaw we have right now is that if you reboot a node that is in "upgrade" state, then crowbar_join will make the node leave that state. Which is probably horribly wrong.

crowbar_framework/app/models/service_object.rb
```diff
@@ -257,20 +257,31 @@ def remove_pending_elements(bc, inst, elements)
     end
   end
 
+  def restore_node(node)
+    node.crowbar['state'] = 'ready'
+    node.crowbar['state_owner'] = ""
```
jsuchome, line 262: Just nit pick, but why use different type of quotes?

crowbar_framework/app/models/service_object.rb
```diff
@@ -257,20 +257,31 @@ def remove_pending_elements(bc, inst, elements)
...
+          restore_node(node)
       end
     end
   end
 
+  def remove_upgrade_state(nodes)
+    nodes.each do |node_name|
+      node = NodeObject.find_node_by_name(node_name)
+      next if node.nil?
+        restore_node(node)
```
jsuchome: Spacing?

nkrinner:
For the record: the upgrade state survives node reboot, and disabled services stay disabled.

jsuchome:
This looks already well. +1 once the correct icon is there

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -338,6 +338,7 @@ en:
         unknown: '%{count} Unknown'
         unready: '%{count} Not Ready'
         pending: '%{count} Pending'
+        upgrade: '%{count} Upgrade State'
```
tboerger, line 341: Just `Upgrade` :)

nkrinner:
Closing, as I won't work on this for now and someone else will take over.

PR #1358, opened 2015-12-16
Implement upgrade state for nodes

jsuchome:
Implement a state to mark nodes ready for upgrades. Also make that
state visible in the UI

(cherry picked from commit fc6254958cdb9e31c5a082c88067cb268033cec1)


crowbar_framework/app/models/service_object.rb
```diff
@@ -257,20 +257,31 @@ def remove_pending_elements(bc, inst, elements)
...
   end
 
+  def remove_upgrade_state(nodes)
+    nodes.each do |node_name|
+      node = NodeObject.find_node_by_name(node_name)
+      next if node.nil?
+        restore_node_to_ready(node)
+    end
+  end
+
```
rhafer, line 279: I am pretty sure with the current state of the updater PR none of the changes here are really needed.

rhafer:
I think we discussed this already but just to not forget about it: BarclampController::valid_transition_states needs to learn about the new "crowbar-upgrade" state.

crowbar_framework/app/models/service_object.rb
jsuchome, line 279: I agree

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -338,6 +338,7 @@ en:
         unknown: '%{count} Unknown'
         unready: '%{count} Not Ready'
         pending: '%{count} Pending'
+        crowbar_upgrade: '%{count} Upgrade'
```
rhafer, line 341: Currently not really important, But this is currently unused. I guess @nkrinner simply forgot to git-add the related changes to nodes_helper.rb to the PR.

jsuchome, line 341: I'll adapt that helper

jsuchome:
@kwwii could you give me the real "led" icon for the upgrade state?

rsalevsky:
@jsuchome We are working on the new nodes view which obsoletes the LEDs. After the PR is merged we just need to adapt the color.

chef/data_bags/crowbar/bc-template-crowbar.json
```diff
@@ -29,19 +29,25 @@
   },
   "deployment": {
     "crowbar": {
-      "crowbar-revision": 0,
+      "crowbar-revision": 1,
```
vuntz: I think this is not the right revision (you want schema-revision, iirc)

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -169,6 +177,17 @@ def create_proposal
     base
   end
 
+  # TODO this code probably could be moved to apply_role
+  def apply_role_pre_chef_call(old_role, role, all_nodes)
+    @logger.debug("crowbar apply_role_pre_chef_call: entering #{all_nodes.inspect}")
+    all_nodes.each do |n|
+      node = NodeObject.find_node_by_name n
+      if node.role? "updater-crowbar-upgrade"
```
vuntz: The role was renamed to crowbar-upgrade, wasn't it?

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,53 @@
...
+    done
+  EOF
+  only_if { node[:platform] == "suse" }
+end
+
+# Disable crowbar-join
+service "crowbar_join" do
+  action :disable
+  only_if { node[:platform] == "suse" }
+end
```
vuntz: Thinking about it: we should probably remove the crowbar_join script from /usr/sbin too.

The reason is that if you move a node to upgrade state, and then reboot / shut down that node, the crowbar_join service will be stopped which will change the state again. The only workaround I can think of is to kill the script.

chef/data_bags/crowbar/bc-template-crowbar.json
```diff
@@ -29,19 +29,25 @@
...
       "crowbar-applied": false,
       "element_states": {
-        "crowbar": [ "all" ]
+        "crowbar": [ "all" ],
+        "crowbar-upgrade": [ "ready", "crowbar-upgrade" ]
```
vuntz: Are you sure it should not contain "applying"?

jsuchome: Not sure yet. Current wat it should work with only "crowbar-upgrade" here. 
That "ready" is for the case when we want to run the "crowbar-upgrade" role for the last time when reverting the node state back to ready. Not sure if this is enough or "applying" is needed, I've not tested it yet

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
jsuchome: Why should it change the state when the node is offline? If nothing is reporting, shouldn;t the last state stay there? I think Nanuk tested rebooting.

vuntz: Because crowbar_join is not stopped, so sysvinit will stop it, which will run crowbar_join --stop, which changes the state of the node :-)

chef/data_bags/crowbar/bc-template-crowbar.json
vuntz: I think in that case you need "applying" (I would put "readying" too just because we usually put the 3 values)

chef/data_bags/crowbar/migrate/crowbar/001_add_crowbar-upgrade-role.rb
```diff
@@ -0,0 +1,16 @@
...
+end
+
+def downgrade ta, td, a, d
+  d['element_states'] = td['element_states']
+  d['element_order'] = td['element_order']
+  d['element_run_list_order'] = td['element_run_list_order']
+  d['config'] = td['config']
+  return a, d
+end
+
```
vuntz: nitpicking: unneeded empty line

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
rhafer: Though it might be worth teaching crowbar_join --stop not to do that when the node in in upgrade state. But just removing it should also work.

crowbar_framework/app/models/service_object.rb
```diff
@@ -257,16 +257,19 @@ def remove_pending_elements(bc, inst, elements)
...
         node = NodeObject.find_node_by_name(node_name)
-        next if node.nil?
-
-        # Nothing to delay so mark them applying.
-        node.crowbar['state'] = 'ready'
-        node.crowbar['state_owner'] = ""
-        node.save
+        next if node.nil? || node.crowbar['state'] == 'crowbar-upgrade'
+          # Nothing to delay so mark them applying.
+          restore_node_to_ready(node)
```
rhafer: This and the `restore_node_to_ready` function are currently not needed. Please remove for now.

chef/data_bags/crowbar/bc-template-crowbar.json
```diff
@@ -31,17 +31,24 @@
     "crowbar": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
+      "schema-revision": 1,
       "element_states": {
-        "crowbar": [ "all" ]
+        "crowbar": [ "all" ],
+        "crowbar-upgrade": [ "ready", "crowbar-upgrade" ]
```
rhafer: I think @vuntz is right that this should be `[ "readying", "ready", "applying", "crowbar-upgrade" ]`

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,53 @@
...
+
+# TODO check for node state: if it is ready, we're likely returning
+# from previous upgrade state and need to re-enable chef-client
+# otherwise do the disabling stuff bellow:
+
+# Disable openstack services
+# We don't know which openstack services are enabled on the node and
+# collecting that information via the attributes provided by chef is
+# is rather complicated. So instead we fall back to a simple bash hack
+
```
rhafer: Because we now have this also on the runlist for some non-upgrade states we need to add a check here for the node state on only do the disable steps when state is crowbar-upgrade. Some like:

```
case node["state"]
  when "crowbar-upgrade"
    do stuff
  when "ready","readying","applying"
    do other stuff
end
```

jsuchome: Yes, it's actually hinted in TODO comment above...

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
jsuchome: I tried to reboot such node, and it stayed in "crowbar-upgrade" when it came up.

jsuchome:
@rhafer so here I have a code executed from controller that should be able to perform both actions, "set to upgrade" and "revert to ready":
https://gist.github.com/jsuchome/953f9ca4a1f356e4af6b
1. Yes, in the end it should be in different functions, or at least do correct action based on some parameter not on current state of nodes
2. Changing node state from the models/crowbar_service.rb (currently in `apply_pre_chef_call`) needs to be removed, or nodes will always end in `crowbar-upgrade` state

However, it does not seem to work well, almost like sometimes the commit function does not apply the (changed) proposal and thus it all fails. I cannot see what is wrong... (there seems to be some queing problem)

rhafer:
Hm, I really wonder how this approach worked on the updater barlcamp (it actually did for me at some point). But ServiceObject::apply_role() (which is called in the end when commiting a proposal) will actually reset the node state to ready in the end (see: https://github.com/crowbar/barclamp-crowbar/blob/release/tex/master/crowbar_framework/app/models/service_object.rb#L1456)

rhafer:
Argh. Actually it was my fault :(

The older version of this PR still had a changed restore_to_ready function. See the hidden comment here: https://github.com/crowbar/barclamp-crowbar/pull/1358#discussion_r48032865

jsuchome:
@rhafer updated. 

Controller action now needs to set extra variable in crowbar_wall: https://gist.github.com/jsuchome/e81f175b9ee2deb9ee08

jsuchome:
I assume gating fails because barclamp-crowbar.spec doesn't expect any migrate scripts (and only one role with certain name)...

rhafer:
As discussed I (more-or-less) quickly implemented the alternative approach for the crowbar-upgrade transition here: https://github.com/rhafer/barclamp-crowbar/tree/upgrade-state-testing (just the last 2 commits on that branch).

I still don't have a strong opinion on which one is better (mine might be somewhat cleaner, but also requires more changes). Feedback is welcome. It shouldn't be too hard to switch between both solutions.

jsuchome:
Regarding gating failure: I am afraid we cannot get gating green, because spec file update of crowbar-barclamp-crowbar (like https://build.suse.de/package/show/home:jsuchome:branches:Devel:Cloud:5:Staging/crowbar-barclamp-crowbar) can only be done after the package has some sources including the migrate directory

jsuchome:
https://bugzilla.suse.com/show_bug.cgi?id=960012 is relevant here, because currently starting chef-client from the recipe fails.

UPDATE: Starting chef-client works on SLE12 nodes...

rhafer:
There is one thing we forgot about: HyperV compute nodes

I think they should be transitioned to the "crowbar-upgrade" state as well, but the chef-recipe should basically be a noop for them. (They need to reinstalled for the upgrade anyway AFAICS)

jsuchome:
Does this PR fail to assign correct state to HyperV compute nodes? Should we remove the constrain from crowbar-upgrade role?

rhafer:
Hm, for now it probably ok to just not apply the role and state-change to HyperV nodes. But that means you need to take care to not push them to the nodes_to_upgrade list. Just check for `node[:platform] == "windows"` an skip those nodes.

Leave the constrain as it is for now.

jsuchome:
Another unresolved part. As already mentioned, rebooting the node does not bring it back to `ready` state, which is correct. However, if I use the Power Actions to do node reboot, it actually changes state to `reboot` and nothing is bringing it back to `crowbar-upgrade`; what now?

I'm thinking of adapting node_object to check for existing `crowbar-upgrade` state before changing the state to `reboot` or `shutdown`... is it a good idea?

Or maybe better one: should we disable Power Action button if node is in `crowbar-upgrade` state (or just every time it is not `ready`)? EDIT: done in last commit

crowbar_framework/app/views/nodes/_show.html.haml
```diff
@@ -18,7 +18,7 @@
       = render :partial => "show_barclamps"
       = render :partial => "show_roles"
 
-  - if @node.bmc_set? or not @node.admin?
+  - if (@node.bmc_set? or not @node.admin?) && @node.state != "crowbar-upgrade"
```
vuntz: Stupid question, but why?

chef/roles/crowbar-upgrade.rb
```diff
@@ -0,0 +1,25 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name "crowbar-upgrade"
+description "Prepare nodes for the cloud upgrade"
```
vuntz: s/cloud/crowbar/

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,67 @@
...
+  # is rather complicated. So instead we fall back to a simple bash hack
+
+  bash "disable_openstack_services" do
+    code <<-EOF
+      for i in /etc/init.d/openstack-* /etc/init.d/openvswitch-switch /etc/init.d/ovs-usurp-config-* /etc/init.d/drbd /etc/init.d/openais;
+      do
+        if test -e $i
+        then
+          initscript=`basename $i`
+          insserv -r $initscript
```
vuntz: In case you change the PR (and if you agree with me): I'd use chkconfig instead of insserv. (just a matter of taste, though)

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,67 @@
...
+    action :disable
+    only_if { node[:platform] == "suse" }
+  end
+
+  # Disable chef-client
+  service "chef-client" do
+    action [:disable, :stop]
+    only_if { node[:platform] == "suse" }
+  end
+else
```
vuntz, line 58: Maybe empty line before `else`

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,67 @@
...
+          insserv -r $initscript
+        fi
+      done
+    EOF
+    only_if { node[:platform] == "suse" }
+  end
+
+  # Disable crowbar-join
+  service "crowbar_join" do
+    action :disable
```
vuntz, line 48: Might be worth adding a comment as to why we don't stop it (as I guess it's not obvious to everyone).

crowbar_framework/app/views/nodes/_show.html.haml
jsuchome: See https://github.com/crowbar/barclamp-crowbar/pull/1358#issuecomment-166855171

If user turns off the node using the Power button, it's state stays as the power action selected.

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,67 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# Check for node state: if it is not 'crowbar-upgrading', we're returning
+# from previous upgrade state and need to re-enable chef-client
+# otherwise do the disabling stuff bellow:
+
+if node["state"] == "crowbar-upgrade"
```
vuntz: I'd feel slightly more comfortable using `node["crowbar_wall"]["crowbar-upgrade"]` for the test. I know it doesn't matter, but it's a gut feeling that it's safer in the long term.

crowbar_framework/app/views/nodes/_show.html.haml
vuntz: Ah, thanks.

vuntz:
So I have minor comments, but overall, this looks good to me!

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -169,6 +182,63 @@ def create_proposal
     base
   end
 
+  def prepare_nodes_for_crowbar_upgrade
+    proposal = ProposalObject.find_proposal('crowbar', 'default')
+
+    # add all nodes new role which prepares them for the upgrade
+    nodes_to_upgrade = []
+    NodeObject.all.each do |node|
+      if node.state == "ready" && !node.admin?
```
vuntz: Oh, actually: the `node.state == "ready"` bit feels wrong to me. If a node is not ready, then we should simply block the upgrade (ie, fail here). The only exceptions are if a node is discovered/discovering.

crowbar_framework/app/models/service_object.rb
```diff
@@ -257,16 +257,19 @@ def remove_pending_elements(bc, inst, elements)
...
+    node.save
+  end
+
   def restore_to_ready(nodes)
     with_lock "BA-LOCK" do
       nodes.each do |node_name|
         node = NodeObject.find_node_by_name(node_name)
-        next if node.nil?
-
+        next if node.nil? || node.crowbar['state'] == 'crowbar-upgrade'
```
rhafer: Some clarifying comment on why the crowbar-upgrade state needs to be handled differently here could be helpful. I already forgot about that once, with really bad results :)

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
jsuchome, line 48: I must admit I also don't know ... :-( 
(this part comes from Nanuk's original commit)

vuntz, line 48: Ah, I can tell you: if we stop, then we change the state of the node to shutdown/reset/something. And we don't want this.

crowbar_framework/app/models/crowbar_service.rb
jsuchome: See last commit

jsuchome:
About the big piechart: I have it always green, even when I try to play with `sliceColors` in js. I probably need to wait for someone who really understands how it works...

rhafer:
> About the big piechart: I have it always green, even when I try to play with sliceColors in js. I probably need to wait for someone who really understands how it works...

How about just leaving all of the related changes out of this PR and worry about that later? I am certainly fine with that.

jsuchome:
Yes, it could be part of the UI PR, I do not mind

rhafer:
+1 but let's see what the CI thinks.

jsuchome:
So it fails, but tempest failures should be unrelated...

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -469,6 +470,11 @@ en:
         raid10: 'RAID 10'
       raid_disks_selected: 'Failed to save %{node}: Not enough disks selected for this RAID level'
 
+  upgrade:
+    upgrade:
+      nodes_not_ready: 'Some nodes are not ready: %{nodes}.
+Fix the state of nodes before proceeding with the upgrade.'
```
vuntz: Can you pick one of the answer at http://stackoverflow.com/questions/10982706/how-to-break-lines-in-yml-localization-files-on-rails for the new line?

vuntz:
+1

jsuchome:
It's green!

nkrinner:
+1

aplanas:
+1

chef/roles/crowbar-upgrade.rb
```diff
@@ -0,0 +1,25 @@
+#
+# Cookbook Name:: updater
```
nkrinner: Nitpick: Wrong cookbook name

jsuchome: Ha!

jsuchome:
@tboerger @rhafer please see last commit, I've tried to address latest problems with JavaScript value mentioned in https://github.com/crowbar/crowbar-core/pull/177

tboerger:
As a general comment, I would prefer new attributes to be written with underscores instead of dashes as this can be used directly as json objects in js and can be used as symbols within ruby or even directly as a class with Hashie::Mash on the rails side.

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,69 @@
...
+# Check for node state: if it is not 'crowbar-upgrading', we're returning
+# from previous upgrade state and need to re-enable chef-client
+# otherwise do the disabling stuff bellow:
+
+if node["crowbar_wall"]["crowbar-upgrade"]
+
+  # Disable openstack services
+  # We don't know which openstack services are enabled on the node and
+  # collecting that information via the attributes provided by chef is
+  # is rather complicated. So instead we fall back to a simple bash hack
```
tboerger: is is :)

chef/cookbooks/crowbar/recipes/crowbar-upgrade.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  bash "disable_openstack_services" do
+    code <<-EOF
+      for i in /etc/init.d/openstack-* /etc/init.d/openvswitch-switch /etc/init.d/ovs-usurp-config-* /etc/init.d/drbd /etc/init.d/openais;
+      do
+        if test -e $i
+        then
+          initscript=`basename $i`
```
tboerger, line 37: Do we prefer `` or $()?

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -169,6 +182,70 @@ def create_proposal
     base
   end
 
+  def prepare_nodes_for_crowbar_upgrade
+    proposal = ProposalObject.find_proposal('crowbar', 'default')
+
+    # To all nodes, add a new role which prepares them for the upgrade
+    nodes_to_upgrade = []
+    all_nodes = NodeObject.all
+    not_ready = all_nodes.select { |n| !n.admin? && n.state != "ready" }.map { |n| n.name }
```
tboerger, line 191: `not_ready = all_nodes.select { |n| !n.admin? && n.state != "ready" }.map(&:name)`

crowbar_framework/app/views/nodes/_show.html.haml
tboerger: This cries for another node object method. Pretty complicated condition for a view.

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -469,6 +470,12 @@ en:
         raid10: 'RAID 10'
       raid_disks_selected: 'Failed to save %{node}: Not enough disks selected for this RAID level'
 
+  upgrade:
+    upgrade:
+      nodes_not_ready: 'Some nodes are not ready: %{nodes}.
```
tboerger: This should be replaced with:

``` yml
  upgrade:
    upgrade:
      nodes_not_ready: |
        Some nodes are not ready: %{nodes}.
        Fix the state of nodes before proceeding with the upgrade.
```

If we really want to have this kind of multi-line messages.

tboerger:
Beside many offenses if we port it to crowbar-core and some comments from my side you get my :+1:

jsuchome:
Hm, so I'd better delete last commit and just replace `crowbar-upgrade` with `crowbar_upgrade` everywhere.

jsuchome:
OK, hopefully finished renaming. Now it will require another test on clean system...

crowbar_framework/app/controllers/barclamp_controller.rb
```diff
@@ -521,6 +521,7 @@ def valid_transition_states
       "applying", "discovered", "discovering", "hardware-installed",
       "hardware-installing", "hardware-updated", "hardware-updating",
       "installed", "installing", "ready", "readying", "recovering",
+      "crowbar_upgrade",
```
vuntz, line 524: I hate being a pain (and should have noticed faster), but the state itself can keep the dash as that's what is being used for other states too.

That being said, most of the other changes to underscore are fine and should be kept.

jsuchome, line 524: I think the node state should be consistent with js we use in UI, so we don't need to add any translation functions...

tboerger, line 524: > I hate being a pain (and should have noticed faster), but the state itself can keep the dash as that's what is being used for other states too.
> 
> That being said, most of the other changes to underscore are fine and should be kept.

I also dislike it there :)

jsuchome, line 524: Didn't you just recently say https://github.com/crowbar/barclamp-crowbar/pull/1358#issuecomment-169682244 ?

The state name is directly used for the state of the group of nodes in nodes_controller, and in relevant js parts. I do not see why it should be different here...

jsuchome:
Tested again, works well (so I did not break it while renaming)

crowbar_framework/app/controllers/barclamp_controller.rb
vuntz, line 524: We can clean this up in a followup PR anyway, imho (the states displayed in leds / the pie are actually a different set of states than the node states, so I would just have an upgrade state there, which would include crowbar-upgrade but possibly other upgrade states).

vuntz, line 524: (btw, I'm not suggesting that you do the followup it as it's not that important for now)

vuntz:
+1

aplanas:
Still +1

PR #1359, opened 2015-12-16
[tex] backport backup part from cloud6

MaximilianMeister:
take only the necessary parts to make a backup

the related PR is https://github.com/crowbar/crowbar-core/pull/144

### DIFF
--- a/crowbar_framework/app/controllers/backup_controller.rb
+++ b/crowbar_framework/app/controllers/backup_controller.rb
@@ -1,0 +1,93 @@
+#
+# Copyright 2015, SUSE LINUX GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class BackupController < ApplicationController
+  #
+  # Backup
+  #
+  # Provides the restful api call for
+  # /utils/backup   POST   Trigger a backup
+  def backup
+    @backup = Backup.new
+    respond_to do |format|
+      if !@backup.nil? && @backup.save
+        format.json { head :ok }
+        format.html { redirect_to backup_path }
+      else
+        msg = I18n.t(".invalid_filename", scope: "backup")
+        format.json { render json: msg }
+        format.html do
+          flash[:alert] = msg
+          redirect_to backup_path
+        end
+      end
+    end
+  end
+
+  #
+  # Download
+  #
+  # Provides the restful api call for
+  # /utils/backup/download/:name/:created_at 	GET 	Download a backup
+  def download
+    @backup = Backup.new(
+      name: params[:name],
+      created_at: params[:created_at]
+    )
+    respond_to do |format|
+      if @backup.path.exist?
+        format.any do
+          send_file(
+            @backup.path,
+            filename: "#{params[:name]}-#{params[:created_at]}.tar.gz"
+          )
+        end
+      else
+        msg = I18n.t(".missing_backup", scope: "backup")
+        format.html do
+          flash[:alert] = msg
+          redirect_to backup_path
+        end
+        format.json { render json: msg }
+      end
+    end
+  end
+
+  #
+  # Delete Backup
+  #
+  # Provides the restful api call for
+  # data-confirm method delete
+  # /utils/backup/delete 	DELETE 	Delete a backup
+  def delete
+    @backup = Backup.new(
+      name: params[:name],
+      created_at: params[:created_at]
+    )
+
+    respond_to do |format|
+      if @backup.valid?
+        @backup.delete
+
+        format.json { head :ok }
+        format.html { redirect_to backup_path }
+      else
+        format.json { render json: @backup.errors }
+        format.html { flash[:alert] = @backup.errors }
+      end
+    end
+  end
+end
--- a/crowbar_framework/app/models/backup.rb
+++ b/crowbar_framework/app/models/backup.rb
@@ -1,0 +1,76 @@
+#
+# Copyright 2015, SUSE LINUX GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class Backup
+  include ActiveModel::Model
+
+  attr_accessor :name, :created_at, :filename, :path
+
+  validates_presence_of :name, :created_at
+  validates :created_at, exclusion: { in: [nil, ""] }
+
+  def initialize(options = {})
+    @name = options.fetch :name, "crowbar_upgrade_data"
+    @created_at = options.fetch :created_at, Time.zone.now.strftime("%Y%m%d-%H%M%S")
+    @filename = "#{@name}-#{@created_at}.tar.gz"
+    @path = Backup.image_dir.join(@filename)
+  end
+
+  def as_json(options: {})
+    result = super
+    result["path"] = path.to_s
+    result
+  end
+
+  def save
+    valid? && create
+  end
+
+  def delete
+    @path.delete
+  end
+
+  def size
+    path.size if path.file?
+  end
+
+  class << self
+    def image_dir
+      if Rails.env.production?
+        Pathname.new("/var/lib/crowbar/backup")
+      else
+        Rails.root.join("storage")
+      end
+    end
+  end
+
+  protected
+
+  def create
+    saved = false
+    dir = Dir.mktmpdir
+
+    Crowbar::Backup::Export.new(dir).export
+    Dir.chdir(dir) do
+      system(
+        "tar czf #{path} *"
+      )
+      saved = true
+    end
+    FileUtils.rm_rf(dir)
+    saved
+  end
+end
--- a/crowbar_framework/config/locales/crowbar/en.yml
+++ b/crowbar_framework/config/locales/crowbar/en.yml
@@ -536,14 +536,23 @@ en:
       two_words_connector: ' and '
       last_word_connector: ' and '
 
+  backup:
+    backup: 'Backup'
+    crowbar: 'Crowbar'
+    openstack: 'OpenStack'
+    crowbar_openstack: 'Crowbar & OpenStack'
+    invalid_filename_format: 'Wrong file format, please use [FILENAME]-[YYYYMMDD]-[HHMMSS].tar.gz'
+    invalid_filename: 'Invalid filename. Please dont use whitespace or special characters'
+    missing_backup: 'Backup does not exist'
+
   # Global
   overview: 'Overview'
   show: 'View'
   all: 'All'
   destroy: 'Remove'
   back: 'Go Back'
-  cancel: 'Cancel'
   edit: 'Edit'
+  delete: 'Delete'
   actions: 'Actions'
   are_you_sure: 'Are you sure?'
   unknown: 'Unknown'
--- a/crowbar_framework/lib/crowbar/backup.rb
+++ b/crowbar_framework/lib/crowbar/backup.rb
@@ -1,0 +1,26 @@
+#
+# Copyright 2011-2013, Dell
+# Copyright 2013-2015, SUSE LINUX GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+module Crowbar
+  class Backup
+    autoload :Base,
+      File.expand_path("../backup/base", __FILE__)
+
+    autoload :Export,
+      File.expand_path("../backup/export", __FILE__)
+  end
+end
--- a/crowbar_framework/lib/crowbar/backup/base.rb
+++ b/crowbar_framework/lib/crowbar/backup/base.rb
@@ -1,0 +1,91 @@
+#
+# Copyright 2011-2013, Dell
+# Copyright 2013-2015, SUSE LINUX GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "logger"
+
+module Crowbar
+  class Backup
+    class Base
+      attr_accessor :logger
+
+      def logger
+        @logger ||= ::Logger.new(STDOUT).tap do |logger|
+          logger.level = ::Logger::DEBUG
+        end
+      end
+
+      class << self
+        def restore_files
+          [
+            [
+              "root/.gnupg",
+              "/root/.gnupg"
+            ],
+            [
+              "root/.ssh",
+              "/root/.ssh"
+            ],
+            [
+              "data",
+              "/var/lib/crowbar"
+            ],
+            [
+              "configs/hosts",
+              "/etc/hosts"
+            ],
+            [
+              "configs/hostname",
+              "/etc/HOSTNAME"
+            ],
+            [
+              "configs/resolv.conf.forwarders",
+              "/etc/resolv.conf"
+            ],
+            [
+              "keys/crowbar-client.pem",
+              "/opt/dell/crowbar_framework/config/client.pem"
+            ],
+            [
+              "keys/webui.pem",
+              "/etc/chef/webui.pem"
+            ],
+            [
+              "keys/chef-client.pem",
+              "/etc/chef/client.pem"
+            ],
+            [
+              "keys/crowbar.install.key",
+              "/etc/crowbar.install.key"
+            ],
+            [
+              "keys/chef-validation.pem",
+              "/etc/chef/validation.pem"
+            ],
+            [
+              "keys/tftp-validation.pem",
+              "/srv/tftpboot/validation.pem"
+            ]
+          ]
+        end
+
+        def export_files
+          restore_files.map(&:reverse)
+        end
+      end
+    end
+  end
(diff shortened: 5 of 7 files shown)


MaximilianMeister:
the idea is that the cloud5 user can make a backup which can be used to be imported into a new cloud6

aplanas:
Isn't a cherry-pick a better option?

MaximilianMeister:
@aplanas not here, we take only the necessary parts to make a backup. and the parent commits have more than what we need here.
and on top of that, the other PR is not merged yet and the commit id's are likely to change...

rsalevsky:
@aplanas And we have different repos so cherry-pick wouldn't work. ;)

crowbar_framework/Gemfile
```diff
@@ -35,6 +35,8 @@ gem "simple-navigation", "~> 3.12"
 gem "simple_navigation_renderers", "~> 1.0"
 gem "sqlite3", "~> 1.3"
 gem "syslogger", "~> 1.6"
+gem "yaml_db", "~> 0.3.0"
```
aplanas: Looks like that gaiting is complaining about this: "`resolve': Could not find gem 'yaml_db (~> 0.3.0) ruby'`"

tboerger:
:+1: but gating fails

MaximilianMeister:
@tboerger that's because @rsalevsky can get the missing dependencies as an update into cloud5 after this is merged IIUC.

MaximilianMeister:
cloud5backup

rhafer:
After applying and restarting crowbar I get this:

```
F, [2015-12-17T14:18:59.453306 #2883] FATAL -- : 
ActionView::Template::Error (translation missing: en.nav.utils.backup):
    11:       = link_to page_title, root_path, :class => "navbar-brand"
    12: 
    13:     .collapse.navbar-collapse.navbar-right
    14:       = render_navigation :expand_all => true, :skip_if_empty => true
```

So it seem some change to `crowbar_framework/config/locales/crowbar/en.yml` is missing from the PR.

rhafer:
Next failure:

```
Errno::ENOENT (No such file or directory @ dir_initialize - /var/lib/crowbar/backup):
  app/models/backup.rb:73:in `all'
  app/controllers/backup_controller.rb:26:in `index'
```

Though I guess it would also be fine if that directory is just part of the rpm package.

rhafer:
After fix those two simple things I get a backup created. But even though the file is called `bla-bla-bla.tar.gz` it's not gzip compressed it a simple (uncompressed) tar file.

crowbar_framework/app/models/backup.rb
```diff
@@ -0,0 +1,140 @@
...
+  def create
+    saved = false
+    dir = Dir.mktmpdir
+
+    Crowbar::Backup::Export.new(dir).export
+    Dir.chdir(dir) do
+      ar = Archive::Compress.new(
+        "#{Backup.image_dir}/#{name}-#{created_at}.tar.gz",
+        type: :tar,
+        compression: :gzip2
```
rhafer: I guess that's either `:bzip2` or `:gzip`, would explain my previous comment.

MaximilianMeister: good catch. thanks I haven't noticed yet

MaximilianMeister:
> So it seem some change to crowbar_framework/config/locales/crowbar/en.yml is missing from the PR.

i think this is in crowbar.yml and after a `Rake assets:precompile` the error may be gone

MaximilianMeister:
> Though I guess it would also be fine if that directory is just part of the rpm package.

~~there is already a submit request from @rsalevsky IIRC~~ merged

rhafer:
> So it seem some change to crowbar_framework/config/locales/crowbar/en.yml is missing from the PR.
> 
> i think this is in crowbar.yml and after a Rake assets:precompile the error may be gone

No, that didn't help here. I really needed to add that key to the en.yml in locales

MaximilianMeister:
> No, that didn't help here. I really needed to add that key to the en.yml in locales

ok I will add it then for now

MaximilianMeister:
@rhafer I hope we have fixed all issues for now, lets wait for the gate

crowbar.yml
```diff
@@ -73,6 +73,9 @@ nav:
     logs:
       order: 10
       route: 'utils_path'
+    backup:
+      order: 20
+      route: 'backup_path'
```
rhafer: I think we should not offer this feature in the utils menu in Tex. It's meant for creating dump that can only be used for upgrading to Master. The dump cannot be restored on Cloud 5, right? If that's true, offering it in the "Utils" menu is creating wrong expectations I think.
The backup for upgrade should be triggerd with the upgrade wizard (whenever that is implemented)

MaximilianMeister: makes sense somehow. but I would rather not drop the link completely. Maybe we can just add a flash message which says sth like "THis Backup Page is meant only to assist the Upgrade to Cloud 6" ?

tboerger: Until we get the wizard I would prefer to keep this link and add a message at the top that clearly states the use case of this page.

crowbar_framework/lib/crowbar/backup/base.rb
```diff
@@ -0,0 +1,95 @@
...
+            [
+              "configs/crowbar.json",
+              "/etc/crowbar/crowbar.json"
+            ],
+            [
+              "configs/hosts",
+              "/etc/hosts"
+            ],
+            [
+              "configs/HOSTNAME",
```
rsalevsky: There is a difference between SLE11SP3 and SLE12SP1, so we should save the file to `configs/hostname`.

crowbar_framework/lib/crowbar/backup/export.rb
```diff
@@ -0,0 +1,177 @@
...
+                file.write("nameserver #{forwarder}")
+              end
+            end
+          else
+            system("sudo", "-i", "cp", "-r", source, data_dir.join(destination).to_s)
+            system("sudo", "-i", "chown", "-R", "crowbar:crowbar", source, data_dir.join(destination).to_s)
+          end
+        end
+
+        data_dir.join("version").open("w") do |file|
```
rsalevsky: This part can be dropped, we have now the meta.

rhafer:
I think it makes sense to squash all of this into a single commit. All of the subsequent commits are basically fixups of the first commit. After that I think this code is fine. (We need to do something about the ci though)

crowbar.yml
vuntz: I've added a task to the upgrade UI card to drop this.

crowbar_framework/lib/crowbar/backup.rb
```diff
@@ -0,0 +1,26 @@
...
+#
+
+module Crowbar
+  class Backup
+    autoload :Base,
+      File.expand_path("../backup/base", __FILE__)
+
+    autoload :Export,
+      File.expand_path("../backup/export", __FILE__)
+  end
```
vuntz, line 25: Stupid question: why is this needed?

crowbar_framework/lib/crowbar/backup/base.rb
```diff
@@ -0,0 +1,95 @@
...
+              "configs/hosts",
+              "/etc/hosts"
+            ],
+            [
+              "configs/hostname",
+              "/etc/HOSTNAME"
+            ],
+            [
+              "configs/network.json",
+              "/etc/crowbar/network.json"
```
vuntz: See https://github.com/crowbar/crowbar-core/pull/144/files#r48938161

crowbar_framework/lib/crowbar/backup/export.rb
```diff
@@ -0,0 +1,173 @@
...
+          "role",
+          "::Chef::Role"
+        )
+      end
+
+      def databags
+        logger.debug "Backing up databags"
+
+        data_dir = workdir.join("knife", "databags")
+        data_dir.mkpath unless data_dir.directory?
```
vuntz: See https://github.com/crowbar/crowbar-core/pull/144/files#r48938462

crowbar_framework/app/views/backup/index.html.haml
```diff
@@ -0,0 +1,52 @@
+.row
```
vuntz: FWIW, I don't think we will need the full UI (as there's no restore for Cloud 5). It's useful for now, but once we get the upgrade UI, I think we can drop that.

crowbar_framework/app/controllers/backup_controller.rb
```diff
@@ -0,0 +1,111 @@
...
+  def backup
+    @backup = Backup.new(
+      name: params[:filename]
+    )
+    respond_to do |format|
+      if !@backup.nil? && @backup.save
+        format.json { head :ok }
+        format.html { redirect_to backup_path }
+      else
+        msg = I18n.t(".invalid_filename", scope: "backup.index")
```
vuntz: See https://github.com/crowbar/crowbar-core/pull/144/files#r48939967

crowbar_framework/app/controllers/backup_controller.rb
```diff
@@ -0,0 +1,111 @@
...
+  # data-confirm method delete
+  # /utils/backup/delete 	DELETE 	Delete a backup
+  def delete
+    @backup = Backup.new(
+      name: params[:name],
+      created_at: params[:created_at]
+    )
+
+    respond_to do |format|
+      if @backup.valid?
```
vuntz, line 82: https://github.com/crowbar/crowbar-core/pull/144/files#r48940093

crowbar_framework/app/models/backup.rb
```diff
@@ -0,0 +1,135 @@
...
+    def image_dir
+      if Rails.env.production?
+        Pathname.new("/var/lib/crowbar/backup")
+      else
+        Rails.root.join("storage")
+      end
+    end
+
+    def filename_time(filename)
+      filename.split(/([\w-]+)-([0-9]{8}-[0-9]{6})/).reject(&:empty?)
```
vuntz: This is not the same regexp as `filename_has_characters` for the filename.

crowbar_framework/app/models/backup.rb
```diff
@@ -0,0 +1,135 @@
...
+
+    def all
+      list = []
+
+      backup_files = image_dir.children.select do |c|
+        c.file? && c.to_s =~ /gz$/
+      end
+
+      backup_files.each do |backup_file|
+        name, created_at = filename_time(backup_file.basename.to_s)
```
vuntz: Shouldn't this be `name, created_at, _ext =` ?

crowbar_framework/app/models/backup.rb
```diff
@@ -0,0 +1,135 @@
...
+
+  attr_accessor :name, :created_at, :filename, :path
+
+  validates_presence_of :name, :created_at
+  validates :created_at, exclusion: { in: [nil, ""] }
+  validate :filename, :filename_has_format
+  validate :filename, :filename_has_characters
+
+  def initialize(options)
+    @name = options.fetch :name, nil
```
vuntz: Shouldn't there be a default instead of `nil` (like `crowbar-backup`)?

crowbar_framework/app/models/backup.rb
```diff
@@ -0,0 +1,135 @@
...
+  end
+
+  def create
+    saved = false
+    dir = Dir.mktmpdir
+
+    Crowbar::Backup::Export.new(dir).export
+    Dir.chdir(dir) do
+      system(
+        "tar czf #{Backup.image_dir}/#{name}-#{created_at}.tar.gz *"
```
vuntz: Why not use `@path` here?

crowbar_framework/app/models/backup.rb
```diff
@@ -0,0 +1,135 @@
...
+    def all
+      list = []
+
+      backup_files = image_dir.children.select do |c|
+        c.file? && c.to_s =~ /gz$/
+      end
+
+      backup_files.each do |backup_file|
+        name, created_at = filename_time(backup_file.basename.to_s)
+        image = new(
```
vuntz: Would name this variable backup, not image.

rhafer:
~~>This part can be dropped, we have now the meta. (https://github.com/crowbar/barclamp-crowbar/pull/1359#discussion_r48841846)~~

~~Hm, now restore on Cloud complains about missing version file.~~

I guess this is a bug in the restore part on master, I'll move my comment there.

crowbar_framework/lib/crowbar/backup/export.rb
```diff
@@ -0,0 +1,173 @@
...
+
+        klass.constantize.list.each do |name, url|
+          logger.debug "Backing up #{component} #{name}"
+
+          record = klass.constantize.load(
+            name
+          ).to_hash.merge(
+            json_class: klass.constantize.to_s
+          )
```
rhafer: See: https://github.com/crowbar/crowbar-core/pull/144/files#r49166530

crowbar_framework/app/models/backup.rb
MaximilianMeister: I have dropped the validations, not needed anymore

crowbar_framework/app/models/backup.rb
MaximilianMeister: I have just not taken care about caching the extension because we don't really need it

crowbar_framework/app/controllers/backup_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+#
+
+class BackupController < ApplicationController
+  #
+  # Backup
+  #
+  # Provides the restful api call for
+  # /utils/backup   POST   Trigger a backup
+  def backup
+    @backup = Backup.new
```
rhafer, line 24: Don't you need to pass parameters here? The model is expecting some.

MaximilianMeister, line 24: not anymore, I have added default values in `initialize`

MaximilianMeister:
I have further dropped some unused code

crowbar_framework/app/controllers/backup_controller.rb
rhafer, line 24: Alternatively you could of course define defaults in the Model's constructor I guess.

crowbar_framework/app/models/backup.rb
```diff
@@ -0,0 +1,76 @@
...
+
+class Backup
+  include ActiveModel::Model
+
+  attr_accessor :name, :created_at, :filename, :path
+
+  validates_presence_of :name, :created_at
+  validates :created_at, exclusion: { in: [nil, ""] }
+
+  def initialize(options)
```
rhafer: Ok, I guess you're missing a `:{}` here. Without that I get: ArgumentError (wrong number of arguments (0 for 1))

crowbar_framework/app/controllers/backup_controller.rb
MaximilianMeister, line 24: > Alternatively you could of course define defaults in the Model's constructor I guess.

yes that's exactly how i did it https://github.com/crowbar/barclamp-crowbar/pull/1359/files#diff-408ac208d889893305de65518aa046a2R27
the second parameter to `fetch` is the fallback default value

crowbar_framework/app/models/backup.rb
MaximilianMeister: hmm strange it worked for me, but i guess you meant adding `options = {}`. okay will do

MaximilianMeister: oh i see in `as_json` we have it like that as well. new Ruby hash syntax. you're right :)

rhafer:
> oh i see in as_json we have it like that as well. new Ruby hash syntax. you're right :)

Actually I was not quite right. Now the download method is broken.

rhafer:
I guess you'd either need to switch to ruby 2 style keyword arguments. Or use the old syntax `options = {}`

MaximilianMeister:
> Actually I was not quite right. Now the download method is broken.

thanks for the hint. fixed

MaximilianMeister:
> I guess you'd either need to switch to ruby 2 style keyword arguments. Or use the old syntax options = {}

it was actually the missing `exist?` method for the backup object, i have just added `@backup.path.exist?`

rhafer:
> it was actually the missing exist? method for the backup object, i have just added @backup.path.exist?

I still get 

```
ArgumentError (unknown keywords: name, created_at):
  app/controllers/backup_controller.rb:46:in `new'
  app/controllers/backup_controller.rb:46:in `download'
```

When calling `curl -u crowbar:crowbar http://localhost:3000/utils/backup/download/crowbar_upgrade_data/20160113-162124 -v --digest`

Am I using the wrong command? It goes away when I change to "options={}" in the model's constructor.

MaximilianMeister:
> Am I using the wrong command? It goes away when I change to "options={}" in the model's constructor.

I was a bit confused as well, `options = {}` is correct here

crowbar_framework/lib/crowbar/backup/base.rb
```diff
@@ -0,0 +1,91 @@
...
+              "root/.gnupg",
+              "/root/.gnupg"
+            ],
+            [
+              "root/.ssh",
+              "/root/.ssh"
+            ],
+            [
+              "data",
+              "/var/lib/crowbar"
```
rhafer, line 44: This leads to all previous backups being included in the current backup. We should avoid that.

crowbar_framework/app/controllers/backup_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+          send_file(
+            @backup.path,
+            filename: "#{params[:name]}-#{params[:created_at]}.tar.gz"
+          )
+        end
+      else
+        msg = I18n.t(".missing_backup", scope: "backup")
+        format.html do
+          flash[:alert] = msg
+          redirect_to backup_path
```
rhafer, line 62: `NameError (undefined local variable or method `backup_path'`

I think this is also true in a number of other places where the backup_path helper is used.

rhafer:
There are still a few issues with the code, especially in the error handling it seems. And depending on the design of the Upgrade wizard it might still require some changes.

But to me it would be ok to merge in the current state so that people can start playing with it more.

So +1 from me and lets address the remaining bit in subsequent pull requests.

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -536,14 +536,23 @@ en:
       two_words_connector: ' and '
       last_word_connector: ' and '
 
+  backup:
+    backup: 'Backup'
+    crowbar: 'Crowbar'
+    openstack: 'OpenStack'
+    crowbar_openstack: 'Crowbar & OpenStack'
```
vuntz, line 543: The strings above are not used, it seems?

crowbar_framework/app/models/backup.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    Crowbar::Backup::Export.new(dir).export
+    Dir.chdir(dir) do
+      system(
+        "tar czf #{path} *"
+      )
+      saved = true
+    end
+    FileUtils.rm_rf(dir)
+    saved
```
vuntz, line 74: Why do we need the `saved` variable?

vuntz:
> But to me it would be ok to merge in the current state so that people can start playing with it more.

Same here, I'm okay merging if there's a followup PR to address the comments.

crowbar_framework/lib/crowbar/backup/export.rb
```diff
@@ -0,0 +1,177 @@
...
+            data_dir.join(destination).open("w") do |file|
+              forwarders.each do |forwarder|
+                file.write("nameserver #{forwarder}")
+              end
+            end
+          else
+            # copy files with higher permissions
+            dest = data_dir.join(destination).to_s
+            system("sudo", "cp", "-a", source, dest)
+            system("sudo", "chown", "-R", "crowbar:crowbar", dest)
```
vuntz, line 110: (just as a reminder from the other PR: this bit will really cause issues later on, we need to keep permissions)

rsalevsky, line 110: Isn't it enough to keep the permissions of the existing files while restoring?

vuntz, line 110: Files might not exist in the new install :/

rhafer:
> Same here, I'm okay merging if there's a followup PR to address the comments.

@vuntz So, I guess I can consider that the second +1?

vuntz:
> @vuntz So, I guess I can consider that the second +1?

Yes

PR #1360, opened 2015-12-17
Add UI for upgrading to Cloud 6

vwallfahrer:
bildschirmfoto vom 2016-01-11 11 35 48


rsalevsky:
Backend part is missing

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -388,6 +389,10 @@ en:
       nics: 'NICs'
       cpu: 'CPU Type'
       hardware: 'Platform'
+    upgrade:
+      next: 'Next'
+      title: 'Upgrade to Cloud 6'
+      upgrade_hint: 'Please make sure you have all the latest maintenance updates installed. To start the upgrade process, click Next.'
```
jsuchome: I think that this section should be independent, not under `nodes:`

crowbar_framework/app/views/upgrade/upgrade.html.haml
```diff
@@ -0,0 +1,9 @@
+.row
+  .col-xs-12
+    %h1.page-header
+      = t("upgrade.title")
```
jsuchome: Just `.title` ? (also others, together with different location in en.yml)

jsuchome:
Good start. Now, we need:
- After user clicks Next, call new action from controller. Code like this should be in such part:
  (but you can use some fake code for now, as this depends on https://github.com/crowbar/barclamp-updater/pull/13)

``` ruby

@service_object = UpdaterService.new logger
proposal = ProposalObject.find_proposal('updater', 'default')

# create new updater proposal if there is not any
if prop.nil?
    proposal_params = {
        "barclamp"=>"updater",
        "name"=>"default",
        "id" => "default",
    }
    @service_object.proposal_create proposal_params
    proposal = ProposalObject.find_proposal('updater', 'default')
end

# add all nodes new role that prepares them for the upgrade
nodes_to_upgrade = []
NodeObject.all.each do |node|
  next if node.admin? || node.state != "ready"
  nodes_to_upgrade.push node.name
end

proposal["deployment"]["updater"]["elements"]["updater-crowbar-upgrade"] = nodes_to_upgrade
proposal.save

answer = @service_object.proposal_commit("default", false, false)
```

(UPDATE: the code above is not relevant any more)
- After the action above is done, another page should be displayed. Now I think there should be a button to revert the previous action ("Bring nodes back to state before upgrade"). 
  And, mostly, a way to get to next step which is apparently exporting of data (point 3 in https://etherpad.nue.suse.com/p/cloud-5-to-6-upgrade and https://github.com/crowbar/barclamp-crowbar/pull/1359). So maybe yet another Next button, redirecting to the Export page?

rsalevsky:
@jsuchome The UI makes no sense for me, is this discussed with Ken?

jsuchome:
@rsalevsky 
1. Not discussed with @kwwii , true
2. Based on https://etherpad.nue.suse.com/p/cloud-5-to-6-upgrade steps 1-2(or 3)
3. Could you elaborate about 'no sense'? It makes sense to me, so there's probably some misunderstanding.

rsalevsky:
Sure, with no sense I mean that there is a lot of blank space on the page for just some buttons. I'm sure we can do it on a single page with maybe some check boxes instead of the next buttons. But this should be discussed with @kwwii. So I don't agree in the way how the process works I just don't like the UI with this blank space and the next buttons.

jsuchome:
I understand it at least in a way of having some preliminary UI that can drive the backend actions. I assume it may be changed later

tboerger:
:-1: because again it's not in any way restful routing...

jsuchome:
@tboerger what exactly is wrong? How could we make it better?

tboerger:
> @tboerger what exactly is wrong? How could we make it better?

Simply the routing, even without the fact that an empty page with a small alert and a next button is pretty crappy UI/UX wise.

``` ruby
resource :upgrade, only: [:show, :update]
```

as route definition and just display a single page instead of the useless `next` button thing.

jsuchome:
I'm fine with this, but I'm not rails expert: @tboerger , @rsalevsky ?

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -469,6 +470,17 @@ en:
...
+  upgrades:
+    show:
+      next: 'Next'
+      title: 'Upgrade to Cloud 6'
+      upgrade_hint: 'Please make sure you have all the latest maintenance updates installed. To start the upgrade process, click Next.'
+    update:
+      title: 'Upgrade to Cloud 6'
+      info: 'Nodes were prepared for upgrade. Click Back to revert this action. Click Next to continue with the Upgrade'
+      reset: 'Reset Upgrade'
+      next: 'Next'
```
rhafer: It would be nice to avoid all these duplicated translations. BTW, for stuff like "Next" we already have global translations.

crowbar_framework/config/routes.rb
```diff
@@ -73,6 +73,11 @@
   get 'utils/upload/:id', :controller=>'support', :action=>'upload', :constraints => { :id => /[^\/]+/ }, :as => 'utils_upload'
   get 'utils/restart/:id(.:format)', :controller=>'support', :action=>'restart', :as => 'restart'
 
+  resource :upgrade, only: [:show, :upgrade] do
+    post 'start'
+    post 'reset'
```
rsalevsky: using `collection` would be nice

kwwii:
According to the etherpad page, this is an extremely complex process which requires user interaction at various steps. The design team should have been involved in this from an early stage. At this point I can give you my opinion on what, if you ask me, is a poorly designed UI and even worse UX...maybe we can make it not suck so much.

PR #1361, opened 2015-12-23
Claim RAID disks when node has full info about them from ohai

jsuchome:
This is an attempt to fix https://bugzilla.suse.com/show_bug.cgi?id=946397

I'm not sure if it is correct. But there is currently indeed a problem that `transition_to_readying` function is called in readying state, when the node still does not have correct information about the disks, thus no RAID disk is actually claimed.

cc @mjura

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -62,8 +62,9 @@ class CrowbarService < ServiceObject
         return [404, "Node not found"]
       end
 
-      if state == "readying"
-        transition_to_readying inst, name, state, node
+      if state == "ready" && !node.admin?
+        # we need to claim raids when we already have ohai data, i.e. after first chef-client run
+        process_raid_claims node
       end
 
       if %w(hardware-installing hardware-updating update).include? state
@@ -310,12 +311,6 @@ class CrowbarService < ServiceObject
   end
 
   protected
-
-  def transition_to_readying(inst, name, state, node = nil)
-    only_unless_admin node do
-      process_raid_claims node
-    end
-  end
 
   def process_raid_claims(node)
     unless node.raid_type == "single"
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -1360,7 +1360,7 @@ class NodeObject < ChefObject
     if device
       crowbar_wall[:claimed_disks] ||= {}
 
-      unless disk_owner(device) == owner
+      if owner.empty? || disk_owner(device) != owner
         return false
       end
 


crowbar_framework/app/models/node_object.rb
```diff
@@ -1360,7 +1360,7 @@ def disk_release(device, owner)
     if device
       crowbar_wall[:claimed_disks] ||= {}
 
-      unless disk_owner(device) == owner
+      if owner.empty? || disk_owner(device) != owner
```
vuntz, line 1363: Isn't the previous test also checking for empty?

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -62,8 +62,8 @@ def transition(inst, name, state)
         return [404, "Node not found"]
       end
 
-      if state == "readying"
-        transition_to_readying inst, name, state, node
+      if state == "ready" && !node.admin?
```
vuntz, line 65: It probably makes sense to explain at least in the commit message (and possibly with a comment here) that we need ohai data, so at least one chef-client run after the install to claim raid disks.

That being said... This means that we try to claim raid disks every time we move to ready, which also means every time we are done with applying a proposal. Not necessarily a big deal, though.

But wouldn't it make more sense to move the claiming to a cookbook (provisioner-base, I guess)? It somehow feels wrong to do it on the webui side.

crowbar_framework/app/models/node_object.rb
jsuchome, line 1363: It was failing without this added test, so maybe there was only a check for nil before

crowbar_framework/app/models/crowbar_service.rb
jsuchome, line 65: I'm not sure how big change do we want for Cloud5 (but the bug is in current code as well).

crowbar_framework/app/models/node_object.rb
vuntz, line 1363: Ah, I read wrong. I guess the issue is that this is called with an empty owner -- this should not happen, so it would be a bug in the code calling this method. Unless we allow forcing a release that way (?).

Do you have more details on when this is happening?

crowbar_framework/app/models/crowbar_service.rb
vuntz, line 65: I think you're right for Cloud 5; take my comment about moving to the cookbook as input when forward-porting the fix ;-)

crowbar_framework/app/models/node_object.rb
jsuchome, line 1363: Well, it is not _currently_ happening - not when `transition_to_readying` is called in `readying` state. At that point , there's no problem with empty owner, but indeed a problem with wrong disk data, so no disk is actually claimed for RAID.
It happens when I move the call of `process_raid_claims` to `ready` state... which again might indicate that such move is not a correct fix for the bug...

vuntz, line 1363: Oh, wait, it should also happen for readying. We unconditionally release, without checking if there was an owner or not.

If I look at the one true source (ie, the similar code in the cookbook), it's fine to call with an empty / nil owner.

Now what I don't get is why this was failing without the changed test :/

mjura:
Ehhh I've tested it and it didn't fix problem

```
      "boot_device": "/dev/disk/by-id/md-name-any:0",
      "claimed_disks": {
        "/dev/vda": {
          "owner": "Boot"
        },
        "/dev/disk/by-id/md-name-any:1": {
          "owner": "OS"
        },
        "/dev/disk/by-id/md-name-any:0": {
          "owner": "OS"
        },
        "/dev/vdc": {
          "owner": "Raid"
        },
        "/dev/vdd": {
          "owner": "Raid"
        }
      },
```

Correct boot device should `/dev/vdc` or `/dev/vdd`

jsuchome:
Thanks for testing. This is what current `process_raid_claims` function does - and as it wasn't invoked correctly before we've probably never met it. 

At least now those disks are claimed somehow.

Now, we may fix this function to add a correct (?) claim string to those disks. Or we may change the place where the boot device is tested so it looks at RAID devices as well.

vuntz:
@jsuchome I think the issue @mjura is pointing out is that /dev/vda should not be claimed (I guess).

rsalevsky:
Cloud 5 (Tex) is EOL since 15th may if this patch is still required please open a new pull request.

PR #1362, opened 2016-01-11
When node is in upgrade state, allow only upgrade specific recipes in runlist

jsuchome:
Back ported from https://github.com/crowbar/crowbar-core/pull/179 cc @rhafer

### DIFF
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -659,7 +659,13 @@ class NodeObject < ChefObject
     crowbar["run_list_map"] = {} if crowbar["run_list_map"].nil?
 
     # Cull by state
-    map = crowbar["run_list_map"].select { |k,v| v["states"].include?("all") or v["states"].include?(self.crowbar['state']) }
+    node_state = self.crowbar["state"]
+    map = crowbar["run_list_map"].select do |k, v|
+      v["states"].include?(node_state) ||
+      # When node is in upgrade state, allow only upgrade specific recipes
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/node_object.rb
```diff
@@ -659,7 +659,12 @@ def rebuild_run_list
     crowbar["run_list_map"] = {} if crowbar["run_list_map"].nil?
 
     # Cull by state
-    map = crowbar["run_list_map"].select { |k,v| v["states"].include?("all") or v["states"].include?(self.crowbar['state']) }
+    map = crowbar["run_list_map"].select do |k, v|
+      v["states"].include?(state) ||
```
vuntz: Actually, I think the move from `self.crowbar['state']` to `state` is wrong: we should never do something depending on `noupdate`, imho.

crowbar_framework/app/models/node_object.rb
```diff
@@ -659,7 +659,12 @@ def rebuild_run_list
...
+      # When node is in upgrade state, allow only upgrade specific recipes
+      (v["states"].include?("all") && !state.include?("upgrade"))
```
vuntz: Isn't the state named crowbar_upgrade?

jsuchome: Yes, that's just for potential other `_upgrade` state. But better to leave the only name we currently have, I guess...

rhafer:
+1

vuntz:
+1

PR #1364, opened 2016-01-20
Add Very basic Upgrade UI for testing

rhafer:
This change should be reverted once the real Upgrade UI is in place.

### DIFF
--- a/crowbar_framework/app/controllers/upgrades_controller.rb
+++ b/crowbar_framework/app/controllers/upgrades_controller.rb
@@ -1,0 +1,43 @@
+#
+# Copyright 2015, SUSE LINUX GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class UpgradesController < ApplicationController
+
+  def show
+  end
+
+  def prepare_nodes
+    service_object = CrowbarService.new(Rails.logger)
+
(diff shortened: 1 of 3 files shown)


jsuchome:
I'm ready to give +1 to this, but still, don't you want to add an action for `revert_nodes_from_crowbar_upgrade` ?

rhafer:
> I'm ready to give +1 to this, but still, don't you want to add an action for revert_nodes_from_crowbar_upgrade ?

Yeah, good point. Will do that.

rhafer:
@jsuchome  Done

crowbar_framework/app/controllers/upgrades_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+class UpgradesController < ApplicationController
+
+  def show
+  end
+
+  def prepare_nodes
+    service_object = CrowbarService.new(Rails.logger)
+
+    # transition nodes to "crowbar_upgrade"
+    service_object.prepare_nodes_for_crowbar_upgrade
```
jsuchome, line 26: Sorry, one more item. Please rescue potential exception here

jsuchome:
I understand there's a need to have something for testing. +1 regardless my comments

crowbar_framework/app/controllers/upgrades_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    @backup = Backup.new
+    @backup.save
+    
+    redirect_to upgrade_path
+  end
+
+  def revert_nodes
+    service_object = CrowbarService.new(Rails.logger)
+
+    # transition nodes to "crowbar_upgrade"
```
jsuchome, line 38: Only in case you plan any further commit: fix this comment.

crowbar_framework/app/controllers/upgrades_controller.rb
rhafer, line 26: I am not sure I want to go down that path (add proper error handling here). What currently happens is the app return a 500 and the details are in the logs. Isn't that enough?

jsuchome, line 26: I'm fine with that for this PR

rsalevsky:
+1

PR #1368, opened 2016-01-26
Revert runlist reduction for crowbar_upgrade state

jsuchome:
introduced by https://github.com/crowbar/baclamp-crowbar/pull/1362

We actually need some core recipes to remain in the runlist every time.

### DIFF
--- a/crowbar_framework/app/models/node_object.rb
+++ b/crowbar_framework/app/models/node_object.rb
@@ -659,12 +659,7 @@ class NodeObject < ChefObject
     crowbar["run_list_map"] = {} if crowbar["run_list_map"].nil?
 
     # Cull by state
-    node_state = self.crowbar["state"]
-    map = crowbar["run_list_map"].select do |k, v|
-      v["states"].include?(node_state) ||
-      # When node is in upgrade state, allow only upgrade specific recipes
-      (v["states"].include?("all") && node_state != "crowbar_upgrade")
-    end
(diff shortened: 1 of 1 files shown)


vuntz:
So that means that all roles that are valid for `all` will stay. Do we have a list of these roles?

rhafer:
> We actually need some core recipes to remain in the runlist every time.

Actually we could do without most of the recipes, they're just a side effect (we need at least ohai::default though). But we need to make sure that all the Proposal Roles (<barclamp>-config-default) (especially those of the core barclamps) stay assigned to the nodes during the upgrade process. Otherwise chef-client on the admin server (e.g. provisioner::update_nodes) will fail.

jsuchome:
I didn't check the roles directly, but this is how list of recipes looks for my node in crowbar_upgrade state:

`Recipes:     barclamp, repos, crowbar-hacks, ohai, kernel-panic, bmc-nat::client, resolver, crowbar::crowbar-db-dump`

rhafer:
> So that means that all roles that are valid for all will stay. Do we have a list of these roles?

It's all "-config-<propsalname>" roles (but those don't have any recipes in their run_list) + crowbar, dns-client and deployer-client.

That resolves to the recipes that @jsuchome just posted.

rhafer:
+1

vuntz:
I wouldn't mind listing the roles / recipes that are left in the commit message. But +1 as it is.

jsuchome:
@vuntz updated commit message

vuntz:
Still +1. Merging as I'm assuming that @rhafer is also still valid (the commit message change doesn't impact code).

PR #1372, opened 2016-01-28
Add wizard for the upgrade task

vwallfahrer:
crowbar


jsuchome:
Please add the actions to the controller. Look at the preliminary UI at https://github.com/crowbar/barclamp-crowbar/pull/1364 for which calls are needed

vwallfahrer:
@jsuchome  okay, I will do that

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -0,0 +1,88 @@
...
+        respond_to do |format|
+          format.any do
+            send_file(
+              @backup.path,
+              filename: @backup.filename
+            )
+          end
+        end
+      else
+        flash[:alert] = "Cannot download image"
```
MaximilianMeister: there we can add some i18n and call it `Failed to create backup` instead

rsalevsky:
Can you include also the steps from https://github.com/crowbar/crowbar-core/pull/221 into your `_nav.html.haml`?

crowbar_framework/app/views/installer/upgrades_nav.html.haml
```diff
@@ -0,0 +1,49 @@
+%ul.nav.nav-stacked.wizard-nav.col-lg-3
```
MaximilianMeister: can we drop this file? as we have `crowbar_framework/app/views/installer/upgrades/_nav.html.haml` which looks almost the same

rsalevsky: Yep, @vwallfahrer you can drop this file.

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -0,0 +1,120 @@
...
+          flash[:alert] = t("installer.upgrades.prepare.failed")
+
+          respond_to do |format|
+            format.html do
+              redirect_to prepare_upgrade_url
+            end
+          end
+        end
+
+        respond_to do |format|
```
rsalevsky, line 33: Put the respond_to directly after `request.post?` and move all the other code inside this block. See https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/controllers/installer/upgrades_controller.rb#L37

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -0,0 +1,120 @@
...
+          service_object = CrowbarService.new(Rails.logger)
+
+          # transition nodes to "crowbar_upgrade"
+          service_object.prepare_nodes_for_crowbar_upgrade
+        rescue
+          flash[:alert] = t("installer.upgrades.prepare.failed")
+
+          respond_to do |format|
+            format.html do
+              redirect_to prepare_upgrade_url
```
rsalevsky: Please show the errors to the user. See https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/controllers/installer/upgrades_controller.rb#L136-L146

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -0,0 +1,120 @@
...
+      begin
+        service_object = CrowbarService.new(Rails.logger)
+
+        # revert nodes from "crowbar_upgrade"
+        service_object.revert_nodes_from_crowbar_upgrade
+      rescue
+        flash[:alert] = t("installer.upgrades.abort.failed")
+        respond_to do |format|
+          format.html do
+            redirect_to prepare_upgrade_url
```
rsalevsky, line 103: Same here, please show the errors.

crowbar_framework/app/views/installer/upgrades/_nav.html.haml
```diff
@@ -0,0 +1,60 @@
...
+        = icon_tag "chevron-right"
+      = t(".download")
+
+  %li{ class: active == :confirm ? "active" : nil }
+    %a
+      .pull-right
+        = icon_tag "chevron-right"
+      = t(".confirm")
+
+  %li{ class: active == :start ? "active" : nil }
```
rsalevsky: just `%li` because this never can reached this is also valid for the flowing li's

crowbar_framework/app/views/installer/upgrades/_nav.html.haml
```diff
@@ -0,0 +1,60 @@
+%ul.nav.nav-stacked.wizard-nav.col-lg-3
+  %li{ class: active == :prepare ? "active" : nil }
+    %a
+      .pull-right
+        = icon_tag "chevron-right"
+      = t(".prepare")
+
+  %li{ class: active == :download ? "active" : nil }
```
rsalevsky, line 8: This step needs should be also added to crowbar-core.

crowbar_framework/app/views/installer/upgrades/_nav.html.haml
```diff
@@ -0,0 +1,60 @@
...
+    %a
+      .pull-right
+        = icon_tag "chevron-right"
+      = t(".download")
+
+  %li{ class: active == :confirm ? "active" : nil }
```
rsalevsky, line 14: This step needs should be also added to crowbar-core.

rsalevsky:
What happens currently after confirming?

crowbar_framework/config/navigation.rb
```diff
@@ -37,6 +37,7 @@
     end
     level1.item :utils, t("nav.utils.title"), utils_path do |level2|
       level2.item :logs, t("nav.utils.logs"), utils_path
+      level2.item :logs, t("nav.utils.upgrade"), prepare_upgrade_path
```
rhafer: `:logs` seems wrong here. Copy'n paste error?

vwallfahrer: Ups... you are right

rhafer:
Some general items (I didn't review the texts, we'll fix those separately I think):
- After clicking on "Begin Upgrade Process" the navigation bar should be hidden if possible, to avoid the user being able to "escape" the upgrade process without running the cleanup
- Is it is possible to show a busy indicator after clicking "Begin Upgrade Process"? The first step will take some time (chef-client run on all nodes).
- The number of steps displayed in the progress bar (8) doesn't match the actual number of steps on the column on the side.
- On the page "Data download" would it be possible to only enable the "Continue ..." Button only if the use clicked on "Download Upgrade Data" before?
- I think we don't need the "Confirm" button on the last page. There is nothing where we could redirect the user from that page. (Currently it redirects to the dashboard. I think we should just stay on that last page).
- The "manual download" link on the last page is wrong. But do we really need it there? Maybe not because we already have that on the page before.

Note: We don't have to fix all of these items in this iteration. But please add them to the Trello card for tracking.

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+          end
+        end
+      end
+
+      respond_to do |format|
+        format.html do
+          redirect_to prepare_upgrade_url
```
rhafer: Shouldn't we redirect to the main dashboard if the upgrade process was successfully aborted?

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+      begin
+        service_object = CrowbarService.new(Rails.logger)
+
+        # revert nodes from "crowbar_upgrade"
+        service_object.revert_nodes_from_crowbar_upgrade
+      rescue => e
+        flash[:alert] = e.message
+        respond_to do |format|
+          format.html do
+            redirect_to prepare_upgrade_url
```
rhafer, line 103: On error, this should probably redirect to itself (`abort_upgrade_url`)

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+#
+
+module Installer
+  class UpgradesController < ApplicationController
+    def prepare
+      if request.post?
+        url = begin
+          service_object = CrowbarService.new(Rails.logger)
+
+          # transition nodes to "crowbar_upgrade"
```
tboerger: IMHO this comment is pretty useless as it's already obvious with the method name.

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+          format.any do
+            send_file(
+              @backup.path,
+              filename: @backup.filename
+            )
+          end
+        end
+      else
+        flash[:alert] = t("installer.upgrades.backup_error")
+        redirect_to download_upgrade_url
```
tboerger:
```
respond_to do |format|
  format.html do
    flash[:alert] = t("installer.upgrades.backup_error")
    redirect_to download_upgrade_url
  end
end
```

crowbar_framework/app/controllers/installer/upgrades_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+      end
+    end
+
+    def abort
+      begin
+        service_object = CrowbarService.new(Rails.logger)
+
+        # revert nodes from "crowbar_upgrade"
```
tboerger: Same as above

crowbar_framework/app/controllers/installer/upgrades_controller.rb
tboerger, line 103: That doesn't make sense as this method is not split into get and post.

crowbar_framework/app/controllers/installer/upgrades_controller.rb
tboerger: Makes sense

crowbar_framework/app/views/installer/upgrades/confirm.html.haml
```diff
@@ -0,0 +1,36 @@
...
+          = render partial: "nav", locals: { active: :confirm }
+
+          .wizard-body.col-lg-9
+            %h2
+              = t(".header")
+
+            %p
+              = t(".description")
+            %p
+              = t(".link")
```
tboerger, line 19: IMHO the link should be a parameter of the translate method.

tboerger:
After fixing my comments you get my :+1:

jsuchome:
This works well already: +1, I vote for handling new/remaining issues in new PR.

(OK, maybe one thing: could you squash some commits into a lesser number?)

(and it needs a rebase anyway)

rhafer:
> OK, maybe one thing: could you squash some commits into a lesser number?

Yes, please!

And then we need to fix the ci I guess. (Seems to be a patch conflict with the branding patch, I can look at that once the commits are squashed)

vwallfahrer:
I rebased the PR and squashed it into one commit. 

And I added the remaining tasks to the trello card.

crowbar_framework/app/views/installer/upgrades/_nav.html.haml
vwallfahrer, line 14: See https://github.com/crowbar/crowbar-core/pull/266

tboerger:
> On the page "Data download" would it be possible to only enable the "Continue ..." Button only if the use clicked on "Download Upgrade Data" before?

I would not do that. Custom javascript for such a more or less useless feature? The user is always forced to download again even if he just reloads the page? This is something I won't do.

> The "manual download" link on the last page is wrong. But do we really need it there? Maybe not because we already have that on the page before.

Where do we get the correct link?

> I think we don't need the "Confirm" button on the last page. There is nothing where we could redirect the user from that page. (Currently it redirects to the dashboard. I think we should just stay on that last page).

I think the confirm button can get removed and we will keep only the back button.

rhafer:
> > The "manual download" link on the last page is wrong. But do we really need it there? Maybe not because we already have that on the page before.
> 
> Where do we get the correct link?

Good question. @MaximilianMeister can you clarify?

rhafer:
I just retriggered the ci with a fixed package. Let's see how far we get.

MaximilianMeister:
> > Where do we get the correct link?
> > Good question. @MaximilianMeister can you clarify?

as we don't have any index like a database record, we don't have any real reusable reference.
when you access `/installer/upgrades/file` via GET you will be sent a new backup, which should be fine as well...
but IMO we don't need the link, as we had the download already before

vwallfahrer:
I removed the link in the follow up PR

rhafer:
CI failed long after deploying and running tempest. This is good enough for this PR.

PR #1379, opened 2016-02-08
Refactor strings in i18n file for upgrade wizard

vwallfahrer:


rsalevsky:
+1

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -558,47 +558,60 @@ en:
...
         blockui_message: 'Aborting upgrade...'
       confirm:
-        header: 'Confirm Download of Upgrade Data'
+        header: 'Reinstall SUSE OpenStack Cloud 6'
         description: |
-          Please reinstall Crowbar with SUSE OpenStack Cloud 6,
-          making sure to use the same configuration used to
-          install SUSE OpenStack Cloud 5.
-        question: 'I have downloaded the upgrade data and I am ready to proceed with the upgrade.'
+          As the next step in the upgrade process you need to redeploy this server
```
jsuchome: Isn't "this server" bit vague here? Shouldn't we rather talk about "Administration Server" as well?

jsuchome:
+1

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -558,47 +558,61 @@ en:
...
+        header_nodes: 'Prepare nodes for upgrade'
+        description_nodes: |
+          As a first step all client nodes of this setup need to be prepared to allow the
+          administration server to be upgraded. Click 'Prepare Nodes for Upgrade' to start
+          that preparation. It is still possible to revert the Upgrade Process after that.
       download:
         cancel: 'Back to selection'
         continue: 'Continue Upgrade Process'
-        abort: 'Abort Upgrade'
+        abort: 'Abort Upgrade and restore client nodes'
```
aplanas: I am not sure if this a title or not

aplanas:
After normalizing the commented sentence: +1

crowbar_framework/config/locales/crowbar/en.yml
vwallfahrer: @aplanas It is a button to abort the upgrade process.

aplanas: so all uppercase then? the thing is that we are mixing both in this sentence

vwallfahrer: You are right, I will change it.

PR #1382, opened 2016-02-09
Add manual download link to upgrade wizard

vwallfahrer:


crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -599,6 +599,7 @@ en:
           will be asked to upload it later on in the process.
           After you downloaded and saved the upgrade data click 'Continue Upgrade Process'.
         blockui_message: 'Aborting Upgrade...'
+        link: 'Manual download link:'
```
vuntz: Why is it "manual"? Or do you mean "Direct download link"?

(tbh, I'm not totally sure how it looks like, maybe a screenshot would help :-))

rhafer:
screen

It looks like this. But the link doesn't work for me.

vuntz:
Thanks. So that'd be "Direct download link". And I'd put it after the button :-)

rsalevsky:
@rhafer The button and the link should point to the same url. Can you check this?

jsuchome:
@vwallfahrer @rsalevsky button leads correctly to `installer/upgrade/file`, but the link to `installer/upgrade/download`

vwallfahrer:
bildschirmfoto vom 2016-02-10 09 41 17

It looks like this now, @kwwii.

vwallfahrer:
I talked to @kwwii and the link has to be above the button.

aplanas:
@vwallfahrer ask @kwwii also about the margins. After that: +1

vuntz:
> I talked to @kwwii and the link has to be above the button.

Ah, sorry.

rsalevsky:
@kwwii Approved this change.

+1

vuntz:
But does the link work? @rhafer said it wasn't working for him.

tboerger:
I would suggest to put it into a text input at the side of the button :)

PR #1384, opened 2016-02-09
Keep openvswitch enabled and disable apache

rhafer:


aplanas:
ahm, not sure about the apache one, is this enabled after the upgrade?

jsuchome:
+1 for apache2, but does disabling openvswitch-switch really hurt? Do you lose the connection after the node reboot?

rhafer:
> ahm, not sure about the apache one, is this enabled after the upgrade

The horizon barclamp will take care of that.

rhafer:
> but does disabling openvswitch-switch really hurt?

I think it depends a lot on the actual network configuration. It just seems to be safer (and more consistent) to not disable it.

jsuchome:
OK then. +1

aplanas:
+1

PR #1385, opened 2016-02-10
Exclude crowbar_upgrade state from exception nodes_not_ready

mjura:


jsuchome:
Why do we need this?

mjura:
> Why do we need this?

For example to re-export once again backup "crowbar_upgrade_data-*.tar.gz". Yesterday, I've exported my backup and went home, today I lost my export and I couldn't re-export it once again, because all nodes were already in crowbar_upgrade states.

jsuchome:
Well, if this should be supported scenario, please explain it in the commit message. And also the code should be adapted a bit to reflect that we do not expect nodes as "ready" only

mjura:
@jsuchome : I've added more comments, but I don't understand what should I change also in code? Could you explain what do you mean?

jsuchome:
Just rename `not_ready` variable

mjura:
> Just rename not_ready variable

`not_ready_for_upgrade` is ok for you?

jsuchome:
Something like that...

mjura:
Ready to go

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -195,10 +195,13 @@ def prepare_nodes_for_crowbar_upgrade
     # To all nodes, add a new role which prepares them for the upgrade
     nodes_to_upgrade = []
     all_nodes = NodeObject.all
-    not_ready = all_nodes.select { |n| !n.admin? && n.state != "ready" }.map { |n| n.name }
+    not_ready_for_upgrade = all_nodes.select { |n| !n.admin? &&
+                                    n.state != "ready" &&
+                                    n.state != "crowbar_upgrade"
```
vuntz: I would have written: `!%w(ready crowbar_upgrade).include?(n.state)`

But it's fine to keep it the current way.

vuntz:
+1

mjura:
@vuntz: I've used your proposal change, thank you

jsuchome:
+1

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -195,10 +195,12 @@ def prepare_nodes_for_crowbar_upgrade
     # To all nodes, add a new role which prepares them for the upgrade
     nodes_to_upgrade = []
     all_nodes = NodeObject.all
-    not_ready = all_nodes.select { |n| !n.admin? && n.state != "ready" }.map { |n| n.name }
+    not_ready_for_upgrade = all_nodes.select { |n| !n.admin? &&
+                                                !%w(ready crowbar_upgrade).include?(n.state)
```
vuntz: We have no hound, but I think that this line is not properly aligned (remove one space, I'd say).

aplanas:
I always will find fun the overuse on `unless` instead of `any` or other functional constructors, but this is Ruby!

+1 for me

crowbar_framework/app/models/crowbar_service.rb
mjura: But then there will be odd number of spaces in line, especially for this, I was checking twice, but if you are fine with removing one space I'm fine too.

MaximilianMeister:
> I always will find fun the overuse on unless instead of any or other functional constructors, but this is Ruby!

maybe it's because `any?` sounds strange, I like `present?` but that is only used in `ActiveRecord` i think

crowbar_framework/app/models/crowbar_service.rb
vuntz: Actually, on multilines, you're not supposed to do `select {}` but `select do end` (according to the style guide)

mjura:
@vuntz : On your wish I've changed implementation style :)

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -194,11 +194,14 @@ def prepare_nodes_for_crowbar_upgrade
 
     # To all nodes, add a new role which prepares them for the upgrade
     nodes_to_upgrade = []
+    not_ready_for_upgrade = []
     all_nodes = NodeObject.all
-    not_ready = all_nodes.select { |n| !n.admin? && n.state != "ready" }.map { |n| n.name }
+    all_nodes.select do |node|
+      not_ready_for_upgrade.push node.name unless !n.admin? && !%w(ready crowbar_upgrade).include?(n.state)
+    end
```
vuntz, line 201: Argh, you lost the `.map { |n| n.name }` bit.

mjura, line 201: But that's way, I am doing `not_ready_for_upgrade.push node.name`. I think this can be replacement

jsuchome, line 201: You're right, `push node.name` is correct, but you're using both `node` variable and `n` (as in `n.admin` and n.state`)

vuntz, line 201: And you're using `select` not `each`, which is very confusing :-)

mjura, line 201: Sorry guys :) too much iterations

mjura:
Fix and ready to go

jsuchome:
+1

mjura:
@jsuchome : Thank you, I've rebased only this change

aplanas:
still +1

crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -194,11 +194,14 @@ def prepare_nodes_for_crowbar_upgrade
 
     # To all nodes, add a new role which prepares them for the upgrade
     nodes_to_upgrade = []
+    not_ready_for_upgrade = []
     all_nodes = NodeObject.all
-    not_ready = all_nodes.select { |n| !n.admin? && n.state != "ready" }.map { |n| n.name }
+    all_nodes.each do |n|
+      not_ready_for_upgrade.push n.name unless !n.admin? && !%w(ready crowbar_upgrade).include?(n.state)
```
vuntz: I think your test is wrong, the correct way would be either:
`if !n.admin? && !%w(ready crowbar_upgrade).include?(n.state)`
or
`unless n.admin? || %w(ready crowbar_upgrade).include?(n.state)`

jsuchome: Oh, true

mjura: Thank you @vuntz

mjura:
Fixed once again

vuntz:
+1

jsuchome:
+1

PR #1387, opened 2016-02-11
Adapt menu strings of upgrade wizard to Cloud6

vwallfahrer:


vuntz:
+1

jsuchome:
+1

vuntz:
No need for gating on that, but before merging, I'd like to get a +1 from @rsalevsky (in case he knows about other strings that need to be sync'ed)

rsalevsky:
+1

PR #1390, opened 2016-02-18
add ActiveRecord migration level to the metadata

MaximilianMeister:
partial backport of https://github.com/crowbar/crowbar-core/pull/343

### DIFF
--- a/crowbar_framework/lib/crowbar/backup/export.rb
+++ b/crowbar_framework/lib/crowbar/backup/export.rb
@@ -116,6 +116,7 @@ module Crowbar
         meta["version"] = ENV["CROWBAR_VERSION"]
         meta["created_at"] = Time.zone.now.to_s
         meta["platform"] = NodeObject.find("role:crowbar").detect(&:admin?).target_platform
+        meta["migration_level"] = ActiveRecord::Migrator.current_version
 
         workdir.join("meta.yml").open("w") do |file|
           file.write(meta.to_yaml)


vuntz:
Hrm, we don't have a db in the backup?

MaximilianMeister:
> Hrm, we don't have a db in the backup?

no, as there is nothing besides maybe the migrations that is worth to backup. and if we would backup the migrations table it would be more effort I think...

vuntz:
@MaximilianMeister sorry, I wasn't clear: is it correct to put a migration level in the backup while we have no db in the backup?

MaximilianMeister:
> is it correct to put a migration level in the backup while we have no db in the backup?

hmm not so sure anymore, in https://github.com/crowbar/crowbar-core/pull/343/files#diff-408ac208d889893305de65518aa046a2R179 it would fall back to `ActiveRecord::Migrator.last_migration.version` if we don't add it here, which I guess should be fine at that point.

if we add it here, it would instead migrate down to that added level and up again to `ActiveRecord::Migrator.last_migration.version` (which would have been the fallback without this patch) which doesn't make too much sense... do you agree?

vuntz:
It feels to me that both should work fine, but I'm not 100% sure. This just needs to be tried. If possible, I'd prefer to avoid further changes here as we already submitted the update to enable 5->6 upgrade.

MaximilianMeister:
>  I'd prefer to avoid further changes here as we already submitted the update to enable 5->6 upgrade.

then let's close this for now

PR #1391, opened 2016-02-18
Refactor update action in nodes controller (bnc#918407)

vwallfahrer:
https://bugzilla.suse.com/show_bug.cgi?id=918407

### DIFF
--- a/crowbar_framework/app/controllers/nodes_controller.rb
+++ b/crowbar_framework/app/controllers/nodes_controller.rb
@@ -338,25 +338,49 @@ class NodesController < ApplicationController
   end
 
   def update
-    unless request.post?
-      raise ActionController::UnknownHttpMethod.new("POST is required to update proposal #{params[:id]}")
-    end
+    respond_to do |format|
+      format.json do
+        unless request.post?
+          render json:  { error: t('nodes.form.post_error', id: params[:id]) }, status: 404 and return
+        end
 
-    get_node_and_network(params[:id] || params[:name])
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/controllers/nodes_controller.rb
```diff
@@ -338,25 +338,49 @@ def edit
   end
 
   def update
-    unless request.post?
-      raise ActionController::UnknownHttpMethod.new("POST is required to update proposal #{params[:id]}")
-    end
+    respond_to do |format|
+      format.json do
+        unless request.post?
```
tboerger, line 343: That part should be dropped and needs to be done through the routes.

rhafer:
Does this really address some bug in bugzilla? If yes, could you please fix the bnc# reference in the title?

crowbar_framework/app/controllers/nodes_controller.rb
MaximilianMeister, line 343: wouldn't it be better to have some nice answer here, instead of an exception on the server side, and a site not found in html?

tboerger, line 343: It would be nicer if this API endpoint just accepts post instead of everything :)

tboerger, line 343: But as this have been like that before let's just keep that for another PR

rhafer:
Given the bug report is against Cloud 6 I am pretty sure that this one was supposed to be done against master first (and then probably backported).

tboerger:
This have also been done for master.

PR #1389, opened 2016-02-17
Refactor dropdown menu for power actions (#bnc955810)

vwallfahrer:
- Dropdown menu collapse after action
- Shows flash message
- Shows blockUI while rebooting

PR #1392, opened 2016-02-24
Refactor dropdown menu for power actions (bsc#955810)

tboerger:
To avoid the situation that it looks like nothing happens we have
removed the AJAX functionality from the power buttons, added a blockui
message, refactored the controller properly to redirect on default html
requests and also refactored the blockui functionality to accept the new
data attribute `data-blockui-click`.

Obsoletes #1389


tboerger:
Lacking activity/interest, closing for now. Feel free to backport the pull request from crowbar/crowbar-core.

PR #1393, opened 2016-02-26
To to fix the provider network parameters of the floating network (bs…

rhafer:
…c#967009)

Because of several bugs (at least bsc#946882 and bsc#955811) we have setups
where the floating network was created with the wrong provider network
settings. While this somehow continued to work and got unnoticed it will create
problems on system upgrading from juno to liberty. Try to adjust the parameters
directly in the database as part of the upgrade.

https://bugzilla.suse.com/show_bug.cgi?id=967009

### DIFF
--- a/chef/cookbooks/crowbar/files/default/crowbar-fix-floating-provider-attributes
+++ b/chef/cookbooks/crowbar/files/default/crowbar-fix-floating-provider-attributes
@@ -1,0 +1,54 @@
+#!/usr/bin/python
+import sqlalchemy
+import sys
+from oslo_config import cfg
+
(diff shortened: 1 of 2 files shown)


rhafer:
This is still not complete. And it proves to be difficult to actually find the right provider parameters and network segment (there might be more networks called "floating") it would be good to also filter for the service tenant id which is hard to get since we lack e.g. access to KeystoneHelper) .  Might be a lot easier if we move this to the neutron barclamp.

rhafer:
Closing in favor of: https://github.com/crowbar/barclamp-neutron/pull/267

PR #1395, opened 2016-05-25
Test if admin node has dns_server role for upgrade

nkrinner:
The upgrade process will fail if the admin node does not have the
dns_server role applied. Abort upgrade with a warning, if this is
not the case. See also
https://bugzilla.suse.com/show_bug.cgi?id=978060

### DIFF
--- a/crowbar_framework/app/models/crowbar_service.rb
+++ b/crowbar_framework/app/models/crowbar_service.rb
@@ -195,9 +195,15 @@ class CrowbarService < ServiceObject
     # To all nodes, add a new role which prepares them for the upgrade
     nodes_to_upgrade = []
     not_ready_for_upgrade = []
+    admin_with_dns_server_role = false
     all_nodes = NodeObject.all
     all_nodes.each do |n|
       not_ready_for_upgrade.push n.name if !n.admin? && !%w(ready crowbar_upgrade).include?(n.state)
+      admin_with_dns_server_role = true if n.admin? && n.role?("dns-server")
+    end
+
+    unless admin_with_dns_server_role
+      raise I18n.t("upgrade.upgrade.admin_missing_dns_server")
     end
 
     unless not_ready_for_upgrade.empty?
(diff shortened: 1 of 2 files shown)


crowbar_framework/app/models/crowbar_service.rb
```diff
@@ -195,9 +195,15 @@ def prepare_nodes_for_crowbar_upgrade
     # To all nodes, add a new role which prepares them for the upgrade
     nodes_to_upgrade = []
     not_ready_for_upgrade = []
+    admin_with_dns_server_role = false
```
jsuchome, line 198: As we are trying to look for incorrect case, it might be better to actually have `admin_without_dns_server_role` (but I do not care that much)

jsuchome:
So, I would be fine with this solution, but I was not sure if we decided having dns-server role on admin is actually a requirement. Didn't we need to check if customer has valid reason to avoid it? (See acceptance criteria)

jsuchome:
+1, based on latest clarification from customer

aplanas:
can this check be done also in the JavaScript side? Does it makes sense?

nkrinner:
@aplanas I don't know how to do it in the javascript part, as we need to access the node data for the check. It is also in line with a similar check that is implemented there.

nkrinner:
Pushed again as i found an artifact in the commit message

jsuchome:
Please rebase

crowbar_framework/config/locales/crowbar/en.yml
```diff
@@ -476,6 +476,9 @@ en:
       nodes_not_ready: | 
         Some nodes are not ready: %{nodes}.
         Fix the state of nodes before proceeding with the upgrade.
+      admin_missing_dns_server: |
+        The admin node does not have the dns-server role applied. During the upgrade process it is 
+        required that the admin node provides this service. Please update your deployment before continuing.
```
vuntz: I'd rephrase to:

"The Administration Server does not have the dns-server role while this is mandatory for the upgrade process. Reapply the DNS proposal with this change before proceeding with the upgrade."

nicolasbock:
+1

jsuchome:
+1

vuntz:
+1 (without the debug commits, of course -- so just for 3e9955f)

nkrinner:
So this currently contains some debug commits (that broke the gating completely)  to fix the travis check. They will be dropped again soon.

nkrinner:
Travis check is unrelated and not required, fix will be submitted later. Merging as we have enough +1

PR #1397, opened 2016-06-01
Fix travis ci checks

nkrinner:
Change Gemfile requirements as active_models_serializers 0.10 release
breakes our call from crowbar_framework/config/initializers/serializer.rb


rsalevsky:
+1

nkrinner:
I did not add that in the commit message, but it might still be useful to know: With the unchanged Gemfile, the Travis CI pulls in a higher version 0.10.0(?) of the active_models_serializers gem than we ship. This higher version is incompatible with our 0.9.\* shipped version. So while code is still ok in our environment, it breaks the Travis CI check.

bmwiedemann:
+1

